PR #20, opened 2018-03-23
Implement Company model

mroetheli:


mroetheli:
Für die neuen models wurden erst model tests geschrieben, die restlichen tests werden zusammen mit der CRUD-Umsetzung geschrieben.

app/controllers/companies_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  private
+    # Use callbacks to share common setup or constraints between actions.
+    def set_company
+      @company = Company.find(params[:id])
+    end
+
+    # Only allow a trusted parameter "white list" through
+# rubocop:disable LineLength
```
psunix: don't ignore this rule :)
just put the arguments on multiple lines.

db/migrate/20180316095930_create_companies.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.string :email
+      t.string :phone
+      t.string :partnermanager
+      t.string :contact_person
+      t.string :email_contact_person
+      t.string :phone_contact_person
+      t.string :crm
+      t.string :level
+      t.string :picture
+      t.boolean :my_company
```
psunix: always use default value for boolean attributes
in this case probably false
t.boolean :my_company, default: false

db/schema.rb
```diff
@@ -70,6 +66,15 @@
     t.index ["person_id"], name: "index_educations_on_person_id"
   end
 
+  create_table "employee_quantities", force: :cascade do |t|
+    t.string "category"
+    t.integer "quantity"
+    t.bigint "company_id"
+    t.datetime "created_at", null: false
```
psunix: isn't there t.timestamps ?

mroetheli: should this be generated automatically?

psunix: yes, the schema.rb is for example updated when you run rake db:migrate
make sure the newest version of schema.rb is checked in.

mroetheli: In the schema.rb all entities have a created_at and updated_at timestamp with datatype t.datetime, there are no t.timestamps. I'm not quite sure, what is false?

psunix: there's nothing wrong ... it's just rails default to use t.timestamps
http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Table.html#method-i-timestamps
you then have only one line :)

PR #34, opened 2018-04-13
Verlinkung Personenprofile auf Ansicht Firmenprofil

mroetheli:


mroetheli:
The branch 'delete confirmation message' is already merged with this branch, check the other branch before this.

PR #40, opened 2018-04-27
Migrate Status to Company

mroetheli:
Assigns company to person according to person.status_id and deletes status_id afterwards. The migration is not applied on this branch.

Apply migration with `rails db:migrate:up VERSION=20180427130110`


db/migrate/20180427130110_migrate_status_to_company.rb
```diff
@@ -0,0 +1,19 @@
+class MigrateStatusToCompany < ActiveRecord::Migration[5.1]
+  def change
```
psunix: never use #change for ireversible migrations

db/migrate/20180427130110_migrate_status_to_company.rb
```diff
@@ -0,0 +1,19 @@
...
+    Person.find_each do |person|
+      case person.status_id
+      when 1
+        person.update!(company_id: 1)
```
psunix: it's always a bad idea relying on any primary keys. 
where are these companies created ?
shouldn't we first create the desired companies ?
how can we identify them ?

mroetheli: We create the company with the seeder. Should we create them via migration?

psunix:
@mroetheli what's the status of these migrations ?

mroetheli:
I'll revise the migrations this week.

mroetheli:
@psunix I've applied the changes.

mroetheli:
I've seen an error in my migration, I'm working on a fix.

mroetheli:
@psunix Now the migration works, but I wasn't able to create a down method which refills all the deleted data. We now have a lot of dependency issues (mentioned by dependency lint), like ember-power-select uses a higher version of a dependency as other dependencies. How can we fix this, I haven't found any working solution.

psunix:
had a strange behavior with the create company migration.
somehow changing the order of attributes fixed it. 
the column offer_comments wasn't created. 
can't explain why, but changing the order fixed it.

```
Caused by:
NoMethodError: undefined method `offer_comment' for #<Company:0x00000002dd08b8>
```
if this error occurs, drop db and recreate it. or just add missing column manually.

psunix:
for me it's fine not having a #down method it the migration.
the current source could would make it incompatible anyways.

psunix:
I do not have any issues with dependencies locally. 
will merge this now. please create a new issue for the dependency problems if the still exist.

PR #66, opened 2018-05-16
Rework last update in company and person

mroetheli:
- company's last update shows updated date when company, offers, locations, employee_quantities gets updatet
- person's last update shows updated date when person, projects, educations, activities, advanced_trainings gets updatet


mroetheli:
In 4. Iteration

PR #120, opened 2019-01-29
added save & new feature, blocked buttons while edit/new

TheWalkingLeek:
ref: #47576


app/models/application_record.rb
```diff
@@ -7,4 +7,12 @@ def start_at_before_finish_at
     return if finish_at.nil? || start_at.nil?
     errors.add(:start_at, 'muss vor "Datum bis" sein') if start_at > finish_at
   end
+
+  def daterange_year_length
+    msg = 'enthält ein zu kurzes Jahr (4 Ziffern)'
```
Robin481: Und was wenn die Zahl zu lang ist? :1234:

PR #160, opened 2019-02-21
Improve people search

mzurbuchen:
People search is now empty if the selected person has been deleted.


Robin481:
Passiert hier noch etwas?

PR #195, opened 2019-04-05
Add searching through CVs, refs: #48610

TheWalkingLeek:


app/models/person.rb
```diff
@@ -61,9 +61,11 @@ class Person < ApplicationRecord
                   against: [
                     :name,
                     :title,
-                    :competences
+                    :competences,
+                    :department
                   ],
                   associated_against: {
+                    skills: :title,
```
janissbinder: roles adden

PR #217, opened 2019-05-21
Add prettier, refs: #48740

TheWalkingLeek:
So this updated a lot of Files because of the formatting...
Important files for reviewing would be package.json, prettierrc, prettierignore and eslintrc.js.
Would also appreciate a quick testing of the whole application just to see if nothing broke while formatting


mzurbuchen:
Looks pretty :tiger:

PR #221, opened 2019-05-27
Fix PeopleSkills Export

TheWalkingLeek:
Sooo I accidentally removed the export backend of people_skills during # 206 when moving the controller and rebasing.
Reimplemented that here :)


Robin481:
:joy: :+1:

PR #262, opened 2019-06-27
Improve skillset table, refs #49566

mzurbuchen:
fixes:  #255


mzurbuchen:
> Clicking on Skills doesn't work anymore. Just throws an error.
> 
> Also, you can ignore the skills name styling so it looks like a link. I did that in a different PR already upside_down_face

It works now :smiley_cat:

PR #316, opened 2019-11-27
Feature/50892 rails 6 upgrade

Thunderguts13:


psunix:
when I execute 'rubocop' without any argument, there's just a few issues shown. we might switch to default config instead of using the .rubocop-must.yml ?
I know, off-topic more or less ... If you think the same way, please remove the must file, adjust the rake ci job.

PR #318, opened 2019-12-09
Feature/50865 radar portfolio sync

Thunderguts13:


db/seeds/development/01_departments.rb
```diff
@@ -0,0 +1,16 @@
...
+                "/dev/tre",
+                "/dev/ruby",
+                "/mid",
+                "/ux",
+                "/zh",
+                "/sys",
+                "/bs",
+                "Funktionsbereiche"]
+
+seeder.seed_departments(departments)
```
psunix: I think you could remove this extra seeder class. It doesn't do a lot.
department_names.each {|n| Department.seed(name: n) }

db/seeds/development/01_person_role_levels.rb
```diff
@@ -0,0 +1,13 @@
...
+
+person_role_levels = ['Keine',
+                       'S1',
+                       'S2',
+                       'S3',
+                       'S4',
+                       'S5',
+                       'S6']
+
+seeder.seed_person_role_levels(person_role_levels)
```
psunix: same as with the department seeder.
maybe use %W syntax for collection.

frontend/app/helpers/get-level.js
```diff
@@ -0,0 +1,15 @@
+import Helper from "@ember/component/helper";
+import { inject as service } from "@ember/service";
+
+export default Helper.extend({
+  store: service(),
+
+  compute(level) {
+    return this.get("store")
+      .findAll("personRoleLevel")
```
psunix: this triggers the person_role_levels api call a second time after loading all candidates for the drop down.
rename helper to: selectedPersonRoleLevel
can't you provide the list of levels as argument here?

PR #328, opened 2020-02-28
Refactor education form and submit

TheWalkingLeek:
That's my progress regarding the refactoring of the education forms.
Basically, every functionality still works and the code is now up to standards of octane. 
I did have to hack the event handling of the escape key (as explained in the comment). 
I also don't completely like the delegation of the submit. For example the submit for the edits is currently in the educations-show component :confused: 
Also the sort-by-year helper had to be divided into a new one, that should be used in the future, and an old one that should get removed once it is not used anymore


frontend/app/utils/sort-by-year-old.js
```diff
@@ -0,0 +1,47 @@
+import { computed, get } from "@ember/object";
+
+// remove this helper after completely moving over to the new version
```
psunix: looks like it would be nice to have some unit tests here. nice task for the students ?

frontend/app/components/activities-edit.js
```diff
@@ -1,5 +1,5 @@
 import Component from "@ember/component";
-import sortByYear from "../utils/sort-by-year";
+import sortByYear from "../utils/sort-by-year-old";
```
psunix: deprecated-sort-by-year ?

PR #343, opened 2020-03-20
Issue 333 cv export notes formatting

### LINKED ISSUE #333: CV-Export notes formatting
When exporting a Person's CV The 'notes' section doesn't format accordingly to the webpage itself.
The goal would be to have the same formatting on an exported Cv as in a Person's profile. (most relevant change is that it doesn't take over line breaks)

Definition of Done:
- The notes section gets formatted the same way in the export as on a Person's profile. 
- There are fitting tests written for this functionality.

Mecanno-man:
Fixes exporter using different format to web app for cv notes, fixes #333

### DIFF
--- a/spec/exporters/odt/cv_spec.rb
+++ b/spec/exporters/odt/cv_spec.rb
@@ -27,9 +27,9 @@ describe Odt::Cv do
       expect(nationalities).to eq('Australien')
     end
 
-    it 'notes formatting' do
-	notes = Odt::Cv.new(people(:bob)).send(:competence_notes_list)[:competence]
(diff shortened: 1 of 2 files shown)


spec/exporters/odt/cv_spec.rb
```diff
@@ -27,5 +27,10 @@
       expect(nationalities).to eq('Australien')
     end
 
+    it 'notes formatting' do
```
TheWalkingLeek: Always formulate the test name by creating a sentence using the `it` keyword. 
In this case, this would probably result in `it 'formats competence notes'`

PR #357, opened 2020-04-03
Feature 335 anonymized cv export

Mecanno-man:
Unsure if this needs frontend tests; seems like everything I changed in the frontend is present untested in other places. Would appreciate feedback.


TheWalkingLeek:
Seems to me like you have a small git mess here. Make sure you rebase correctly on master so you, for example, don't try to merge the same commit a second time. You can use the commits section on this page to see what it's trying to merge

app/controllers/people_controller.rb
```diff
@@ -40,10 +40,13 @@ def person
     @person ||= Person.find(params[:person_id])
   end
 
-  def export
-    odt_file = Odt::Cv.new(entry).export
+  # rubocop:disable Metrics/LineLength
```
TheWalkingLeek, line 43: Try to avoid disabling rubocop like this...

app/controllers/people_controller.rb
```diff
@@ -40,10 +40,13 @@ def person
...
+  def export(anon = params[:anon].presence || 'false')
+    odt_file = Odt::Cv.new(entry).export(anon != 'false')
```
TheWalkingLeek, line 45: Refactor this so you pass the params to the constructor... This way it's also expendable and you don't need to do these weird checks.
Also, if you do set a default value for a method, avoid using operators and other objects. It should be a fallback, what you're doing is assigning a variable.

app/controllers/people_controller.rb
```diff
@@ -40,10 +40,13 @@ def person
...
     send_data odt_file.generate,
               type: 'application/vnd.oasis.opendocument.text',
-              disposition: content_disposition('attachment', filename(entry.name, 'cv'))
+              disposition: content_disposition('attachment',
+                                               anon != 'false' ? 'cv.odt' : filename(entry.name, 'cv'))
```
TheWalkingLeek, line 49: The name of the file should be anonymized_cv.odt according to the DoD.
Also refactor this filename to a variable. You need to make ruby code readable by naming things and creating sentences using this naming. If you need to declare a variable to do so, do it. Juggling around with ternary and other things only makes it more complicated and less readable.

app/exporters/odt/cv.rb
```diff
@@ -8,10 +8,11 @@ def initialize(person)
       @person = person
     end
 
-    def export
-      ODFReport::Report.new('lib/templates/cv_template.odt') do |r|
-        insert_general_sections(r)
-        insert_personalien(r)
+    # rubocop:disable Metrics/LineLength
+    def export(anon = true)
+      ODFReport::Report.new(anon ? 'lib/templates/cv_template_anon.odt' : 'lib/templates/cv_template.odt') do |r|
```
TheWalkingLeek, line 13: Here you can use a variable again. Also, the path stays the same... just concat the path string with the correct filename

app/exporters/odt/cv.rb
```diff
@@ -8,10 +8,11 @@ def initialize(person)
       @person = person
```
TheWalkingLeek, line 8: Here you can then add the params as an instance variable

app/exporters/odt/cv.rb
```diff
@@ -8,10 +8,11 @@ def initialize(person)
...
 
-    def export
-      ODFReport::Report.new('lib/templates/cv_template.odt') do |r|
-        insert_general_sections(r)
-        insert_personalien(r)
+    # rubocop:disable Metrics/LineLength
+    def export(anon = true)
+      ODFReport::Report.new(anon ? 'lib/templates/cv_template_anon.odt' : 'lib/templates/cv_template.odt') do |r|
+        insert_general_sections(r, anon)
+        insert_personalien(r, anon)
```
TheWalkingLeek, line 15: Now you won't need to pass the anon

app/exporters/odt/cv.rb
```diff
@@ -39,16 +41,21 @@ def insert_general_sections(report)
...
-      report.add_field(:nationalities, nationalities)
-      report.add_field(:email, person.email)
-      report.add_image(:profile_picture, person.picture.path) if person.picture.file.present?
+      unless anon
+        report.add_field(:birthdate, Date.parse(person.birthdate.to_s)
+                                         .strftime('%d.%m.%Y'))
+      end
+      report.add_field(:nationalities, nationalities) unless anon
+      report.add_field(:email, person.email) unless anon
+      report.add_image(:profile_picture, person.picture.path) unless anon || person.picture.file.blank?
```
TheWalkingLeek, line 56: why use unless anon as a block first and then use it in guard close after as well? Just put everything into the unless anon block

frontend/app/components/person-cv-export.js
```diff
@@ -0,0 +1,32 @@
...
+  },
+
+  actions: {
+    startExport(personId, e) {
+      e.preventDefault();
+
+      let url =
+        `/api/people/` +
+        this.get("router.currentURL").split("/")[2] +
+        `.odt?anon=false`; //TODO: This is ugly
```
TheWalkingLeek, line 19: Yes it is. You even have the personId as a variable

frontend/app/components/person-cv-export.js
```diff
@@ -0,0 +1,32 @@
...
+      this.get("download").file(url);
+    },
+
+    startAnonymizedExport(personId, e) {
+      e.preventDefault();
+      let url =
+        `/api/people/` +
+        this.get("router.currentURL").split("/")[2] +
+        `.odt?anon=true`; //TODO: This is ugly
```
TheWalkingLeek, line 28: same here

frontend/app/templates/components/person-cv-export.hbs
```diff
@@ -0,0 +1,25 @@
+<div class="modal fade" id="person-cv-export" tabindex="-1" role="dialog" aria-labelledby="person-cv-export-label" aria-hidden="true" data-backdrop="false">
+  <div class="modal-dialog modal-dialog-centered modal-lg" role="document">
```
TheWalkingLeek, line 2: remove the modal-lg class. This way it is way too big

app/controllers/people_controller.rb
```diff
@@ -41,9 +41,11 @@ def person
   end
 
   def export
-    odt_file = Odt::Cv.new(entry).export
+    anon = params[:anon].presence || 'false'
+    odt_file = Odt::Cv.new(entry, anon).export
+    filename = anon != 'false' ? 'anonymized_cv.odt' : filename(entry.name, 'cv')
```
TheWalkingLeek, line 46: That's not what i meant, just pass all params like `odt_file = Odt::Cv.new(entry, params).export`.
Also why use such a weird statement? Can't you just use `params[:anon].present?`? booleans as strings are probably always a bad idea

app/exporters/odt/cv.rb
```diff
@@ -4,12 +4,15 @@
 module Odt
   class Cv
 
-    def initialize(person)
+    def initialize(person, anon)
       @person = person
+      @anon = (anon != 'false')
```
TheWalkingLeek, line 9: Then change this to `@params = params`.
You can now define a new method like this:
```
def anon
@params[:anon].present?
end
```
and use that method instead of `@anon`

app/exporters/odt/cv.rb
```diff
@@ -39,16 +43,21 @@ def insert_general_sections(report)
       report.add_field(:version, '1.0')
       report.add_field(:comment, 'Aktuelle Ausgabe')
     end
-    # rubocop:enable Metrics/AbcSize
 
+    # rubocop:enable Metrics/AbcSize
     # rubocop:disable Metrics/AbcSize
```
TheWalkingLeek, line 48: Don't need to enable it if you disable it on the next line? :D

app/exporters/odt/cv.rb
```diff
@@ -39,16 +43,21 @@ def insert_general_sections(report)
...
-      report.add_field(:nationalities, nationalities)
-      report.add_field(:email, person.email)
-      report.add_image(:profile_picture, person.picture.path) if person.picture.file.present?
+      unless @anon
+        report.add_field(:birthdate, Date.parse(person.birthdate.to_s)
+                                         .strftime('%d.%m.%Y'))
+        report.add_field(:nationalities, nationalities)
+        report.add_field(:email, person.email)
+      end
+      report.add_image(:profile_picture, person.picture.path) unless @anon || person.picture.file.blank?
```
TheWalkingLeek, line 58: You can put that into the block as well. Just change it back to `if person.picture.file.blank?`

app/controllers/people_controller.rb
Mecanno-man: but wouldn't checking if anon is present lead to unexpected behavior when calling `anon=false`?

TheWalkingLeek: Are you doing that? I thought it would either send with the param being true or not present at all.
But if you do then yes it would, change it `anon == 'true'` then though to be more readable

app/exporters/odt/cv.rb
```diff
@@ -4,12 +4,19 @@
...
 
-    def initialize(person)
+    def initialize(person, params)
       @person = person
+      @params = params
     end
 
+    def anon?
+      @params[:anon].present? && @params[:anon] == 'true'
+    end
```
TheWalkingLeek, line 14: Move this below the private keyword so it's a private method
You can also refactor the statement to `@params[:anon].presence == 'true'`

Issue #334, opened 2020-03-06
'Stationen' month field semi-obligatory


Robin481:
As our backoffice needs the exact duration of a person's stay at a company, the field 'month' for 'Stationen' would need to be obligatory. 
The obvious problem is that a Person filling out their Cv might not always know the specific months they stayed at a company. (or atleast not right at the moment of filling it out)

Therefore the solution we will approach is: 
- Making the field 'semi-obligatory', meaning when you first submit a 'Station' without the month filled in you will get a warning saying something along the lines of 'Please fill in a month, if you want to save anyway press **here'.** Then on pressing **here** the Station will get saved even without a month. 
- This should atleast remind users that the specific month is an important information and should hopefully raise the amount of people filling this field out. 

Definition of Done:
- When first saving a 'Station' without the month filled in a warning will be displayed. But one can still save without a month by pressing on the **here** button in the warning text.
- The warning doesnt show when there are months provided. (Be aware that the "to" field has a "until today" option. This should also count as correctly filled out and not cause an error)
- There are fitting tests written for this functionality.

Robin481:
![month (3)](https://user-images.githubusercontent.com/32168247/76609955-85679500-6518-11ea-922d-9dc31d11ec60.png)

PR #358, opened 2020-04-05
Feature #334 'Stationen' month field semi-obligatory

KnownInter:
Added feature #334

### DIFF
--- a/frontend/app/components/activity-new.js
+++ b/frontend/app/components/activity-new.js
@@ -34,20 +34,8 @@ export default Component.extend(EKMixin, {
     context.sendAction("done", true);
   },
 
-  addMissingMonthWarning(newActivity) {
-    if (
-      !newActivity.monthFrom ||
-      (!newActivity.monthTo && newActivity.yearTo)
-    ) {
-      document.getElementById("missing-month-warning").style.display = "block";
-      return true;
-    } else if (
-      document.getElementById("missing-month-warning").style.display == "block"
-    ) {
-      document.getElementById("missing-month-warning").style.display = "none";
-      return false;
-    }
-    return false;
+  showMissingMonthWarning() {
+    this.$("#missing-month-warning").css("display", "block");
   },
 
   actions: {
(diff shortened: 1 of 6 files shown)


frontend/app/components/activity-new.js
```diff
@@ -34,14 +34,31 @@ export default Component.extend(EKMixin, {
     context.sendAction("done", true);
   },
 
+  addMissingMonthWarning(newActivity) {
```
TheWalkingLeek, line 37: I would divide the boolean return and the display.style stuff into separate methods... The returned boolean is a check on the record and the style stuff is specific to this component...

Refactor this to something like `showMissingMonthWarning()` and only do the style stuff here

frontend/app/components/activity-new.js
```diff
@@ -34,14 +34,31 @@ export default Component.extend(EKMixin, {
...
+    if (
+      !newActivity.monthFrom ||
+      (!newActivity.monthTo && newActivity.yearTo)
+    ) {
```
TheWalkingLeek, line 41: Put this check into the record as a function... Maybe even into daterange-model.js and give it a smart name.
That way it is reusable and only bound to the model.

frontend/app/components/activity-new.js
```diff
@@ -34,14 +34,31 @@ export default Component.extend(EKMixin, {
...
+      document.getElementById("missing-month-warning").style.display = "block";
```
TheWalkingLeek, line 42: Use jquery here. I know it probably shows a deprecation or something but we use jquery in other places as well and in that case, rather be consistent than hack around :)

frontend/app/components/activity-new.js
```diff
@@ -34,14 +34,31 @@ export default Component.extend(EKMixin, {
...
   actions: {
     abortNew(event) {
       event.preventDefault();
       this.sendAction("done", false);
     },
 
-    submit(newActivity, initNew, event) {
+    submit(newActivity, initNew, monthNeeded, event) {
       event.preventDefault();
+      if (monthNeeded && this.addMissingMonthWarning(newActivity)) return;
```
TheWalkingLeek, line 61: Here you would then refactor this to something like
```
if (monthNeeded && newActivity.areMonthsGiven()) {
showMissingMonthWarning();
return;
}
```
Maybe think of some better names :)

frontend/app/models/daterange-model.js
```diff
@@ -9,5 +9,10 @@ export default DS.Model.extend({
   isYearValid(year) {
     if (year.length == 4 && year > 0) return true;
     return false;
+  },
+
+  missesMonth() {
+    if (!this.monthFrom || (!this.monthTo && this.yearTo)) return true;
+    return false;
```
TheWalkingLeek, line 16: `return !this.monthFrom || (!this.monthTo && this.yearTo)` should do the trick. Also maybe rename to missesMonths as in plural :)

Issue #337, opened 2020-03-06
Sorting of skills when adding a new one


Robin481:
When adding a new Skill under: Profile -> Skills -> new-Skills, the suggested skills are sorted by Uppercase and Lowercase, meaning all lowercase skills will appear at the end of the list ignoring the alphabetical order.

Definiton of done:
- The Skills are only sorted by alphabetical order (not upper- lowercase)
- There is a fitting test written for this functionality.

dovaya:
The reason why the sort function *orderByTitleCaseInsensitive* in the *skills/frontend/app/components/people-skill-new.js* file tests if the skill titles are undefined is because whenever the modal to add a new People-Skill is opened and closed/aborted without actually adding a new People-Skill, both an empty skill and an empty People-Skill are created (->empty records with every property set to *undefined*). This bug was present before I changed any code. -> needs more investigation

PR #359, opened 2020-04-05
Issue #337 sorting_of_skills_when_adding_a_new_one

dovaya:
changed the dropdown in the new-Skill-modal to list the available skills alphabetically and case-insensitively, and added a test to check the feature

### DIFF
--- a/frontend/app/components/people-skill-new.js
+++ b/frontend/app/components/people-skill-new.js
@@ -39,16 +39,16 @@ export default Component.extend({
   }),
 
   orderByTitleCaseInsensitive(firstSkill, secondSkill) {
-    let firstTitle =
-      firstSkill.get("title") == undefined
-        ? undefined
-        : firstSkill.get("title").toLowerCase();
-    let secondTitle =
-      secondSkill.get("title") == undefined
-        ? undefined
-        : secondSkill.get("title").toLowerCase();
+    let firstTitle = firstSkill.title && firstSkill.title.toLowerCase();
+    let secondTitle = secondSkill.title && secondSkill.title.toLowerCase();
 
-    return firstTitle < secondTitle ? -1 : 1;
+    if (firstTitle < secondTitle) {
+      return -1;
+    }
+    if (secondTitle > firstTitle) {
+      return 1;
+    }
+    return 0;
   },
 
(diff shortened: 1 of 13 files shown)


frontend/app/components/people-skill-new.js
```diff
@@ -34,10 +34,23 @@ export default Component.extend({
...
+    let firstTitle =
+      firstSkill.get("title") == undefined
+        ? undefined
+        : firstSkill.get("title").toLowerCase();
+    let secondTitle =
+      secondSkill.get("title") == undefined
+        ? undefined
+        : secondSkill.get("title").toLowerCase();
+
+    return firstTitle < secondTitle ? -1 : 1;
```
TheWalkingLeek, line 51: Replace this with three return statements
```
  if (firstTitle < secondTitle) {
    return -1;
  }
  if (secondTitle > firstTitle) {
    return 1;
  }
  return 0
```
This way, you also cover the case of them being the same string. It's also just like in the example on: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

frontend/app/components/people-skill-new.js
```diff
@@ -34,10 +34,23 @@ export default Component.extend({
...
-      return skills.sort((a, b) => (a.get("title") < b.get("title") ? -1 : 1));
+      return skills.sort((a, b) => this.orderByTitleCaseInsensitive(a, b));
     });
   }),
 
+  orderByTitleCaseInsensitive(firstSkill, secondSkill) {
+    let firstTitle =
+      firstSkill.get("title") == undefined
+        ? undefined
+        : firstSkill.get("title").toLowerCase();
```
TheWalkingLeek, line 45: You can refactor this to `firstTitle = firstSkill.title && firstSkill.title.toLowerCase();`
This works because at first it tries to call title on firstSkill. If title is `undefined` it will not even try toLowerCase() because it's a falsey value (and thus the && doesn't bother to try the second option because it's already falsey no matter what). If somethings there, it will call the toLowerCase() and return that. That's just JS "magic" that you gotta get used to :)

Also keep in mind that in JS you should ALWAYS compare using `===`. It's not needed anymore after the refactoring but keep it in mind. It's a small error but can lead to big bugs. :upside_down_face:

Issue #336, opened 2020-03-06
Skill-Search minimum amount of experience


Robin481:
When searching for Skills we want to be able to narrow the results by adding the "minimum amount of experience" parameter.

In the Skill-Search view there should be an experience-selector, in the same style as in the Skill-Rating of a user, that allows the person searching to define a minimum amount of experience to search for. 

Per Default the experience selection is not active and it displays the skills unfiltered. Only when a specific experience level is selected the search will filter accordingly. 

Definition of done: 
- A second parameter "minimum amount of experience" can be searched for in the skill-search field.
- A reset button is next to the slider that resets the search parameter
- There are fitting tests written for this functionality.

Robin481:
Mockups:
![notfiltered](https://user-images.githubusercontent.com/32168247/76760114-59087e80-678d-11ea-803e-956e29ea880f.png)
![filtered2](https://user-images.githubusercontent.com/32168247/76760117-5b6ad880-678d-11ea-88b9-714af5dadbf0.png)

PR #361, opened 2020-04-05
Feature #336 Skill Search minimum amount of experience

glowseeker:
Tests are incomplete

also, unfortunately for some reason, my VS Code suddenly showed +1k image modified files, so I manually picked out the ones I remember editing so I might've missed some.


TheWalkingLeek:
Please don't assign stuff to me that is not yet finished. Feel free to use the PR as draft to track your progress but i really just want to review code that is truly finished :)

frontend/app/controllers/skill-search.js
```diff
@@ -19,13 +26,42 @@ export default class SkillSearchController extends Controller {
...
+
+  get levelName() {
+    const levelNames = [
+      "Nicht bewertet",
+      "Trainee",
+      "Junior",
+      "Professional",
+      "Senior",
+      "Expert"
+    ];
```
TheWalkingLeek, line 51: Refactor this into the people-skills.js model 
Just like here: https://github.com/puzzle/skills/blob/d54cb8202395bd6203d834601f98cef8fee04964/frontend/app/models/person.js#L33

frontend/app/controllers/skill-search.js
```diff
@@ -11,6 +11,13 @@ export default class SkillSearchController extends Controller {
   @service
   router;
 
+  currentSkillId = 0;
+
+  init() {
+    super.init(...arguments);
+    this.set("levelValue", 1);
```
TheWalkingLeek, line 18: this.levelValue = 1
This is kind of confusing because we just switched to Ember Octane. this.set and this.get are still old syntax. Just use the variables like any other instance variable :)

frontend/app/controllers/skill-search.js
```diff
@@ -19,13 +26,42 @@ export default class SkillSearchController extends Controller {
...
+    this.currentSkillId = skillId;
     return skillId ? this.get("store").peekRecord("skill", skillId) : null;
   }
 
+  @action
+  updateSelection() {
+    this.get("router").transitionTo({
+      queryParams: {
+        skill_id: this.currentSkillId,
+        level: this.get("levelValue")
```
TheWalkingLeek, line 38: this.levelValue

frontend/app/controllers/skill-search.js
```diff
@@ -19,13 +26,42 @@ export default class SkillSearchController extends Controller {
   @computed("model")
   get selectedSkill() {
     const skillId = this.router.currentRoute.queryParams.skill_id;
+    this.currentSkillId = skillId;
     return skillId ? this.get("store").peekRecord("skill", skillId) : null;
   }
 
+  @action
+  updateSelection() {
```
TheWalkingLeek, line 34: Change this to `filterPeopleSkills(skillId, level)`. Then you can call this action from both the dropdown and the slider.

frontend/app/controllers/skill-search.js
```diff
@@ -19,13 +26,42 @@ export default class SkillSearchController extends Controller {
...
+    return levelNames[this.get("levelValue")];
+  }
+
   @action
   setSkill(skill) {
     this.get("router").transitionTo({
-      queryParams: { skill_id: skill.get("id") }
+      queryParams: { skill_id: skill.get("id"), level: this.get("levelValue") }
     });
   }
```
TheWalkingLeek, line 60: You can now remove this action

app/controllers/people_skills_controller.rb
```diff
@@ -28,7 +27,14 @@ def fetch_entries
...
+      filter_entries(people_skills)
+    end
+  end
+
+  def filter_entries(people_skills)
+    people_skills = people_skills.where(skill_id: params[:skill_id])
+    if params.key?(:level)
+      people_skills.where('level >= ?', params[:level])
     end
   end
```
TheWalkingLeek, line 39: Overloading methods in ruby is rather rare...
Just pass the params[:level] to the PeopleSkillsFilter on line 26 and handle the filter logic in there
Same as with the rated param. Create a new private method `filter_by_level` and filter there

spec/controllers/people_skills_controller_spec.rb
```diff
@@ -25,16 +25,33 @@
...
         json_object_includes_keys(skill_attrs, keys)
       end
+
+      it 'only returns above a level' do
+        keys = %w[person_id skill_id level interest certificate core_competence]
+      
+        process :index, method: :get, params: { type: 'Skill', skill_id: rails.id, level: '2'}
+
+        skills = json['data']
+        expect(skills.count).to eq(1)
```
TheWalkingLeek, line 44: Also check if the skill name and level are as expected

spec/controllers/people_skills_controller_spec.rb
```diff
@@ -25,16 +25,33 @@
...
+      end
+
+      it 'returns level itself' do
+        keys = %w[person_id skill_id level interest certificate core_competence]
+      
+        process :index, method: :get, params: { type: 'Skill', skill_id: rails.id, level: '3'}
+
+        skills = json['data']
+        expect(skills.count).to eq(1)
```
TheWalkingLeek, line 53: same here

frontend/app/controllers/skill-search.js
glowseeker: Did you mean people-skill.js that is in the same folder? people-skills.js doesn't exist :P

frontend/app/controllers/skill-search.js
Mecanno-man: Not quite sure how this is supposed to work: the action setSkill requires a skill, not an id - this skill is only accessible from the dropdown and isn't saved anywhere else, as such the slider can't access it. We would need at least a second action to set the skill id.

Mecanno-man: disregard; refractored to be called via observers due to issues with the slider.

frontend/app/controllers/skill-search.js
TheWalkingLeek: right, I meant exactly that ;)

frontend/app/controllers/skill-search.js
Mecanno-man: This change won't work due to the switch to observers, and the octane observer syntax doesn't work in @classic

app/domain/people_skills_filter.rb
```diff
@@ -1,15 +1,20 @@
...
+    @level = level
   end
 
   def scope
-    filter_by_rated
+    if level.nil?
+      filter_by_rated
+    else
+      filter_by_level
+    end
```
TheWalkingLeek, line 17: I wouldn't do the conditional here. Just call both methods. But you need to pass the entries you received from the `filter_by_rated` method to the `filter_by_level` method. Thus, change both method declarations to having the entries as parameters and then work with the return value.

app/domain/people_skills_filter.rb
```diff
@@ -23,4 +28,8 @@ def filter_by_rated
     end
     entries
   end
+
+  def filter_by_level
+    entries.where('level >= ?', level)
+  end
```
TheWalkingLeek, line 34: Here you can just use
`return entries if level.nil?`

The idea of the scope method is to just filter everything using the parameters passed on initialization. If you do an if else statement, it won't run all the filters. It's not the scope methods concern whether or not there even is a level value. It just want's to run all filters over it. When the filter does not have the required value, it just returns the received entries

frontend/app/controllers/skill-search.js
TheWalkingLeek: i see :)

app/domain/people_skills_filter.rb
```diff
@@ -23,4 +24,8 @@ def filter_by_rated
     end
     entries
   end
+
+  def filter_by_level(entries)
+    level.nil? ? entries : entries.where('level >= ?', level)
+  end
```
TheWalkingLeek, line 30: I would write it like this:
```
 def filter_by_level(entries)
    return entries if level.nil?

    entries.where('level >= ?', level)
  end
```
Sure you write more lines but it is more readable this way. In this ternary you either return the value of a variable or the result of a db query. Those are completely different "actions" if you get what I mean. It's not really one or the other (for which you would use a ternary) but a completely different thing.
It's not always about the length of the method. You need to make your code as readable as possible. If you combine the readability with ruby/rails methods, naming and CoC for smart, short, readable code, you're on top!

Issue #345, opened 2020-03-27
Frontend validations


Robin481:
Currently we have almost no Frontend Input Validations. All data is validated in the backend. 
To make it easier to do some fast on-the-fly validations we want to implement these validations in the frontend as well. (This would for example allow us to colour some fields red during a wrong input.)

First we will implement Frontend Validations for the person-edit component and from there on see how much work it would take for the rest of the components. 

Use a well supported and smart addon that runs on the most recent version of EmberJs and has an active community behind it. (Compare your findings and discuss it with @TheWalkingLeek before finalizing the decision)

Also consider (https://github.com/puzzle/skills/issues/347) in your decision, maybe you will be able to find an Addon that alreads supports this. 

Definition of done: 
- The person-edit component now has frontend validation
- Tests :factory:

PR #367, opened 2020-04-27
Feature #345 Frontend validations

KnownInter:
Added feature #345

### DIFF
--- a/frontend/app/components/validation-date-input.js
+++ b/frontend/app/components/validation-date-input.js
@@ -1,11 +1,13 @@
 import Component from "@ember/component";
+import { action } from "@ember/object";
+import classic from "ember-classic-decorator";
 
-export default Component.extend({
-  actions: {
-    setBirthdate(value) {
-      this.get("setDirty")();
-      this.get("update")(value);
-      this.set("input", value);
-    }
+@classic
+export default class ValidationDateInput extends Component {
+  @action
+  setBirthdate(value) {
+    this.setDirty();
+    this.update(value);
+    this.set("input", value);
   }
-});
+}
--- a/frontend/app/components/validation-date-input.js
+++ b/frontend/app/components/validation-date-input.js
@@ -1,0 +1,11 @@
+import Component from "@ember/component";
+
+export default Component.extend({
+  actions: {
+    setBirthdate(value) {
+      this.get("setDirty")();
+      this.get("update")(value);
+      this.set("input", value);
+    }
+  }
+});
--- a/frontend/app/templates/components/person-edit.hbs
+++ b/frontend/app/templates/components/person-edit.hbs
@@ -84,12 +84,7 @@
             <table class="table-person">
               <tbody>
                 <tr>
-                  <td>
-                    <label class="label-date form-field--label" for="date_location">Geburtsdatum</label>
-                    <div id="date_location" class="birthdate_pikaday">
-                      <PikadayInput @onSelection={{action "setBirthdate"}} @value={{this.person.birthdate}} @useUTC={{true}} />
-                    </div>
-                  </td>
+                  <td>{{personChangeset.input label="Geburtsdatum" name="birthdate" renderComponent=(component "validation-date-input" input=this.person.birthdate) errorComponent=(component "validation-error-component")}}</td>
                 </tr>
                 <tr>
                   <td>{{personChangeset.input label="Wohnort (Stadt)" name="location" errorComponent=(component "validation-error-component")}}</td>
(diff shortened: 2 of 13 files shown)


frontend/app/components/validation-date-input.js
```diff
@@ -0,0 +1,11 @@
+import Component from "@ember/component";
+
+export default Component.extend({
+  actions: {
+    setBirthdate(value) {
+      this.get("setDirty")();
+      this.get("update")(value);
```
TheWalkingLeek, line 7: I have never seen something like this with two parantheses. Aren't there any better ways?

frontend/app/components/validation-date-input.js
```diff
@@ -0,0 +1,11 @@
+import Component from "@ember/component";
+
+export default Component.extend({
```
TheWalkingLeek, line 3: Change this to a glimmer component!

frontend/app/templates/components/person-edit.hbs
```diff
@@ -84,15 +84,10 @@
             <table class="table-person">
               <tbody>
                 <tr>
-                  <td>
-                    <label class="label-date form-field--label" for="date_location">Geburtsdatum</label>
-                    <div id="date_location" class="birthdate_pikaday">
-                      <PikadayInput @onSelection={{action "setBirthdate"}} @value={{this.person.birthdate}} @useUTC={{true}} />
-                    </div>
-                  </td>
+                  <td>{{personChangeset.input label="Geburtsdatum" name="birthdate" renderComponent=(component "validation-date-input" input=this.person.birthdate) errorComponent=(component "validation-error-component")}}</td>
```
TheWalkingLeek, line 87: This validation didn't work for me...

frontend/app/templates/components/person-edit.hbs
```diff
@@ -151,13 +146,11 @@
           </div>
         </div>
         <br>
-        <button {{action "submit" personChangeset}} role="button" id="submit-button" class="btn btn-primary">
-          Speichern
-        </button>
+        {{personChangeset.submit label="Speichern" id="submit-button"}}
```
TheWalkingLeek, line 149: Could you also add the bootstrap btn classes back to this? :)

frontend/app/components/validation-date-input.js
KnownInter: I didn't find any other ways, so probably no.

frontend/app/components/validation-date-input.js
KnownInter: Changing it to a glimmer component breaks it. Is it necessary?

frontend/app/templates/components/person-edit.hbs
KnownInter: It seems fine to me and also passed the test. What exactly didn't work?

frontend/app/templates/components/person-edit.hbs
KnownInter: It should've the bootstrap btn classes since I'm using the bootstrap theme for the validation addon.

KnownInter:
> Honestly I'm very (positively) surprised this worked! 😄
> I was working on a frontend validation using the same addons and it just didn't work for me... Looks awesome here though. A bit sad it doesn't work for power select fields but that's fine for now.
> 
> Just a few nitpicks. Could you also wrap the submit action inside a check for the isValid of the record so we don't submit it unless it is valid?
> Good work though!

It should work for power select fields by creating a custom component like I did for the birthdate input, but I deemed it as unnecessary since there is no way to have an invalid value.
And wrapping the submit action inside a check for the validity seems needless to me because it shouldn't perform the action as long as the form is invalid.

frontend/app/templates/components/person-edit.hbs
TheWalkingLeek: That one's on me... don't know why it didn't work. Works now though, sorry about that :sweat_smile:

frontend/app/templates/components/person-edit.hbs
TheWalkingLeek: I must've not executed yarn install or something. You're obviously right, the styling's completely fine.

frontend/app/components/validation-date-input.js
TheWalkingLeek: I just took this into my own hands, hope you don't mind.
You are right, you're currently relying on methods with this component by the classic component.
In that situation, to not lose a lot of time I decided to change it to the class syntax and add the classic decorator. Best to update this once the addon gets updated

PR #369, opened 2020-05-07
Feature/341 cv search jump to

DominikFischli:
Implemented the Jump-To with a component.

Functionality:
- The component tries to jump to the found query string; If the query string is not located anywhere in the searchable attributes, it times out and tries again after a set amount of time.
- A static MAX_TIMEOUT value of 2000 (2 seconds) has been specified after which the function aborts.

Caveats:
- No Tests: I couldn't figure out how to properly test this functionality, also due to the Error below.
- `Error: You attempted to schedule an action in a quee (afterRender) for a method that doesn't exist (backburner.js:429)`
For some Reason, `scheduleOnce("afterRender", context, jumpOrRetry(context));` does sometimes throw this Error; Debugging this shows that the schedule function of backburner.js is indeed called without a given method to execute.
Possibly this is caused by `Ember.run.scheduleOnce` being called from within `Ember.run.later`. 
Despite the Error, the method succeeds in its purpose.


TheWalkingLeek:
I've texted you directly about this issue :sweat_smile:

PR #371, opened 2020-05-08
Readded company model & functionality

DominikFischli:
Company is again visible in the Person Detail View and a company can be selected when creating a new person.

Seeders have been readded.

The migration has been rewritten - as before, the unused tables location, offers, employee_quantities stay removed but the company table stays with all unused columns removed.


frontend/app/components/person-edit.js
```diff
@@ -196,6 +201,8 @@ export default ApplicationComponent.extend(EKMixin, {
         person.rollbackAttributes();
       }
 
+      this.set("person.company", this.get("callBackCompany"));
```
TheWalkingLeek, line 204: This doesn't seem to be working, if you abort the edit it still stays the new value

Mecanno-man:
Note that if both this Pull Request and [Pull Request 374](https://github.com/puzzle/skills/pull/374) are merged, tests will fail again. To prevent this, add a company to the new test persona in /spec/fixtures/people.yml before merging the second.

DominikFischli:
I present you the two-in-one bug solver: As it turns out, departments had the same issue with still displaying the changed department despite aborting. (Department had not even logic in place to deal with the situation the way company did)

The reason for the bug: 
`this.callBackCompany = this.get("person.company");` linked instead of loaded the company.
Therefore, this.callBackCompany and this.person.company referenced the same company at all times, instead of callBackCompany referencing the previously selected company.

TheWalkingLeek:
Nice, I just quickly changed the seed to seed_once to not get twin records in the database :)

PR #374, opened 2020-05-13
Feature342 skill search and

Mecanno-man:
Note that if both this Pull Request and [Pull Request 371](https://github.com/puzzle/skills/pull/371) are merged, tests will fail again. To prevent this, add a company to the new test persona in /spec/fixtures/people.yml before merging the second.


app/controllers/people_skills_controller.rb
```diff
@@ -23,18 +23,22 @@ def fetch_entries
...
   end
 
+  def fetch_person_entries(base)
+    people_skills = PeopleSkillsFilter.new(base, params[:rated]).scope
+    people_skills.where(person_id: params[:person_id])
+  end
+
+  def fetch_skill_entries(base)
+    PeopleSkillsFilter.new(base, params[:rated], params[:level], params[:skill_id]).scope
+  end
```
TheWalkingLeek, line 40: Just call the `PeopleSkillsFilter.new(...).scope` once like on line 39 and then if there is a :person_id param, just do the where on the result

app/domain/people_skills_filter.rb
```diff
@@ -1,16 +1,17 @@
 # frozen_string_literal: true
 
 class PeopleSkillsFilter
-  attr_reader :entries, :rated, :level
+  attr_reader :entries, :rated, :level, :skill_id
 
-  def initialize(entries, rated, level = nil)
+  def initialize(entries, rated, level = nil, skill_id = nil)
```
TheWalkingLeek, line 6: I just realized, assigning nil as default value does absolutely nothing. If you pass nothing, the value is nil. You can just remove the default value assignment

app/domain/people_skills_filter.rb
```diff
@@ -1,16 +1,17 @@
...
     @entries = entries
     @rated = rated
-    @level = level
+    @level = level.nil? ? nil : level.to_s.split(',')
+    @skill_id = skill_id
```
TheWalkingLeek, line 10: rename them to `@levels` and `@skill_ids`

app/domain/people_skills_filter.rb
```diff
@@ -25,7 +26,33 @@ def filter_by_rated
     entries
   end
 
-  def filter_by_level(entries)
-    level.nil? ? entries : entries.where('level >= ?', level)
+  def filter_by_level(entries, skills)
+    return entries if level.nil?
+    skill_ids = skills.to_s.split(',')
```
TheWalkingLeek, line 31: split this at the same location as the levels

app/domain/people_skills_filter.rb
```diff
@@ -25,7 +26,33 @@ def filter_by_rated
...
+    if skill_ids.length == @level.length
+      return get_filter_results(entries, skill_ids)
+    else
+      raise ArgumentError, 'Amount of Skill_ids and levels provided do not match'
+    end
+  end
+
+  def get_filter_results(entries, skill_ids)
```
TheWalkingLeek, line 39: rename to `level_filter_results`. Also, you don't need to pass the skill_ids here since it's an instance variable
`get_filter_results` is very general

frontend/app/controllers/skill-search.js
```diff
@@ -3,7 +3,6 @@ import { action, computed } from "@ember/object";
 import { observes } from "@ember-decorators/object";
 import { inject as service } from "@ember/service";
 import Controller from "@ember/controller";
-import PeopleSkill from "../models/people-skill";
 
 @classic
 export default class SkillSearchController extends Controller {
```
TheWalkingLeek, line 8: Okay so about this. You've probably noticed there is a looooot of repeated code here, that's not the idea.

My idea would be:
- The SkillSearchController has an array. There is one Object in the array by default. 
- The object has two attributes: level, skill
- Every time you press the "Skill hinzufügen" Button, an object gets added to the array
- In the skill-search.hbs: loop over the array and for every object, show a component.
- That component has the slider and powerselect for one level and one skill. These are the ones you pass when iterating over the array
- If a value gets changed, bubble an action to the SkillSearchController.
- That action gets the values of all the filter components, these should be in the array. And then create the backend call with all the data inside the Controller

app/domain/people_skills_filter.rb
```diff
@@ -25,7 +26,33 @@ def filter_by_rated
...
+    person_ids = []
+    entries.where(buildstring(skill_ids)).group(:person_id).count.each_entry do |key, value|
+      person_ids.push(key) if value == skill_ids.length
+    end
+    entries.where(person_id: person_ids, skill_id: skill_ids)
```
TheWalkingLeek, line 44: How about you do it like so:
```
person_ids = PeopleSkill.where(skill_id: 18, level: 3).or(PeopleSkill.where(skill_id: 12, level: 1)).group_by(&:person_id).transform_values(&:count).map { |id, count| id if count == skill_ids.length }.compact
```
This is obviously crazy for one line. Split the results to several well named variables. But this way you can get all the person_ids you need from one query and don't need some crazy buildstring method.
Of course, you have to do one `or(PeopleSkill.where())` for every array entry.

app/domain/people_skills_filter.rb
Mecanno-man: If these are not assigned something by default, I get an argument error when calling it without a level and skill_id.

app/domain/people_skills_filter.rb
```diff
@@ -25,7 +26,33 @@ def filter_by_rated
...
+
+  def buildstring(skill_ids)
+    result = +''
+    skill_ids.each_with_index do |skill, index|
+      result.concat('(skill_id=')
+            .concat(skill.to_s)
+            .concat(' and level >=')
+            .concat(@level[index].to_s)
+            .concat(') or ')
+    end
```
TheWalkingLeek, line 55: You can also try to build a string like this for the OR statement just two wishes:
1) change the method name to something that makes genuine sense and makes it readable
2) escape the params: https://guides.rubyonrails.org/active_record_querying.html#array-conditions

This code could easily allow sql injection which is really dangerous!

TheWalkingLeek:
One more note, take a look at the mockup provided in the Issue. The UI doesn't really look like that. Try to fix that styling as well, thanks :)

frontend/tests/integration/components/people-skill-show-test.js
```diff
@@ -21,7 +21,6 @@ module("Integration | Component | people-skill-show", function(hooks) {
 
     let text = this.$().text();
 
-    assert.ok(text.includes("Rails"));
```
TheWalkingLeek, line 24: Why was this just removed?

app/domain/people_skills_filter.rb
TheWalkingLeek: I see, didn't know that. In that case it's fine :)

spec/fixtures/people.yml
```diff
@@ -58,3 +58,15 @@ ken:
   competence_notes: LaTex\n Puppet\n Bash
   email: kenthemaster@testi.com
   department: ux
+
+lain:
+  birthdate: 1998-07-06 01:15:00
+  location: Tokyo
+  marital_status: <%= Person.marital_statuses[:single] %>
+  name: Lain Iwakura
```
Robin481, line 66: Please use made-up names in our fixtures and not names of anime characters.

frontend/tests/integration/components/people-skill-show-test.js
Mecanno-man: The Skill title is now an optional feature and not loaded in the way that this component is tested

Robin481:
New PR: https://github.com/puzzle/skills/pull/532

PR #400, opened 2021-02-09
Feature/398 selectable branch

Robin481:
TODO:
- Automatically pre-select first Branch-Adress in Dropdown


app/controllers/branch_adresses_controller.rb
```diff
@@ -0,0 +1,8 @@
+class BranchAdressesController < CrudController
+
+  private
+
+  def fetch_entries
+    BranchAdress.all
+  end
```
TheWalkingLeek, line 7: Do you even need this? Should call `model_scope.list` inside ListController anyways

frontend/app/models/branch-adress.js
```diff
@@ -0,0 +1,7 @@
+import DS from "ember-data";
+
+export default DS.Model.extend({
+  short_name: DS.attr("string"),
+  adress_information: DS.attr("string"),
```
TheWalkingLeek, line 5: use `-` instead of `_`

frontend/app/components/person-cv-export.js
```diff
@@ -1,27 +1,65 @@
...
+import { isBlank } from "@ember/utils";
 
 export default class PersonCvExport extends Component {
   @service download;
   @service router;
+  @service store;
 
-  init() {
-    super.init(...arguments);
+  locations = this.store.findAll("branch_adress");
```
TheWalkingLeek, line 11: `branch-address`

frontend/app/components/person-cv-export.js
```diff
@@ -1,27 +1,65 @@
...
+    let blurredEl = e.relatedTarget;
+    if (isBlank(blurredEl)) {
+      return false;
+    }
+    return !blurredEl.classList.contains("ember-power-select-search-input");
+  }
+
+  @action
+  setLocation(location) {
+    Ember.set(this, "selectedLocation", location);
```
TheWalkingLeek, line 39: `this.selectedLocation = location`
Should be working...

frontend/app/templates/components/person-cv-export.hbs
```diff
@@ -7,6 +7,15 @@
           <span aria-hidden="true">&times;</span>
         </button>
       </div>
+        <div class="modal-body">
+          Niederlassung (für Footer):
+          <div class="container-fluid">
+            <PowerSelect @selected={{this.selectedLocation}} @options={{this.locations}} @onFocus={{action "handleFocus" }}
+                            @onBlur={{action "handleBlur" }} @onChange={{action "setLocation"}} as |location|>
```
TheWalkingLeek, line 14: instead of `{{action "handleFocus"}}` use `{{this.handleFocus}}`

spec/models/branch_adress_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe BranchAdress, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
TheWalkingLeek, line 4: remove that? :P

TheWalkingLeek (changes requested):

frontend/app/components/person-cv-export.js
```diff
@@ -1,27 +1,65 @@
...
 
 export default class PersonCvExport extends Component {
   @service download;
   @service router;
+  @service store;
 
-  init() {
-    super.init(...arguments);
+  locations = this.store.findAll("branch_adress");
+  selectedLocation = this.locations.get("firstObject");
```
TheWalkingLeek, line 12: this might not be working because the findAll on locations is async.
You're not waiting for this. Either set the selectedLocation inside a `then` block. Or make the locations `@tracked` and change the selectedLocation to `get selectedLocation()` https://ember-learn.github.io/ember-octane-vs-classic-cheat-sheet/#component-properties__args

frontend/app/components/person-cv-export.js
Robin481: I had that before and sadly it didn't work. I'm not sure why but with Ember.set it works as expected.

frontend/app/models/branch-adress.js
Robin481: With `-` we get a build error so I doubt that that is correct.
I've changed it to camelCase for now as it seems more inline with the rest of the tool.

frontend/app/templates/components/person-cv-export.hbs
Robin481: Good call

app/controllers/branch_adresses_controller.rb
Robin481: Yes and no :wink: 
I didn't define a `list` scope as it didn't seem necessary to me as these records should be ordered by id anyway. As I didn't have a `list` scope defined this would not have worked.
However, SQL can be a bit wonky as to which order it returns records in. Therefore it makes sense to define the `list` scope anyway and send the records ordere by `ID` just to be on the sure side.
So now that I have a `list` scope your suggestion is of course correct. :partying_face:

app/exporters/odt/cv.rb
```diff
@@ -11,9 +11,10 @@ def initialize(person, params)
 
     # rubocop:disable Metrics/MethodLength
     def export
-      template_name = anon? ? 'cv_template_anon.odt' : 'cv_template.odt'
+      template_name = "cv_template#{'_de' if location.country == 'DE'}#{'_anon' if anon?}.odt"
```
TheWalkingLeek, line 14: Maybe just create a variable or method for the suffix. This looks kinda all over the place

TheWalkingLeek (approved):
looking good

PR #403, opened 2021-02-09
Feature/402 GitHub actions

Robin481:


.github/workflows/rails.yml
```diff
@@ -0,0 +1,45 @@
+name: Rails
+on: [push, pull_request]
+jobs:
+  Rails-Tests:
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:11
```
psunix, line 9: we're using postgres 9.5 in production. 
maybe use same version as in prod here?

.github/workflows/rails.yml
```diff
@@ -0,0 +1,45 @@
...
+    - uses: actions/checkout@v2
+    - uses: ruby/setup-ruby@v1
+      with:
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+
+    - name: Setup test database
+      env:
+        RAILS_ENV: test
+        PGHOST: localhost
+        POSTGRES_DB: rails_github_actions_test
```
psunix, line 31: is there a reason not to go with default values here? e.g. db name skills_test ?

.github/workflows/rails.yml
```diff
@@ -0,0 +1,45 @@
...
+        POSTGRES_USER: rails_github_actions
+        POSTGRES_PASSWORD: postgres
+      run: |
+        cp config/database.ci.yml config/database.yml
```
psunix, line 35: is it not possible to go with default datbase.yml here ?

.github/workflows/rails.yml
```diff
@@ -0,0 +1,45 @@
...
+        bundle exec rake db:create db:migrate
```
psunix, line 36: is this required? db:migrate is executed when calling rails ci
see lib/tasks/ci.rake

psunix (commented):
I don't really know github actions yet ... but added some thoughts

.github/workflows/rails.yml
Robin481: Yes totally agree

.github/workflows/rails.yml
Robin481: Yes, but the CI Task also tries to execute the frontend tests which are not intended to be used in this GitHub Action. We would probably have to split up the rake tasks if we wanted to use the ci rake task. 
I think having the commands in the workflow file makes it simpler to follow anyway.

.github/workflows/rails.yml
Robin481: Yes it does indeed work :)

Issue #417, opened 2021-03-08
CV Export - Only Freitext


Robin481:
**Feedback from Andreas Gurtner:**
When exporting a CV, the information about all the Skills can be unnecessary. For most use cases the "Freitext" Feld "Notizen" is the only thing that is needed in the exported CV. This is due to the fact that usually the information in there would be adapted anyway according to the company the offer goes to. 

**Solution**
Let the user chose whether to include the Skills in the exported CV or only to include the "Notizen" Feld.
When pressing the "export" button there should be an option whether to include the Skills or not. 
Depending on which option the user chose the Skills should be included in the exported CV or not.
The default case should be to include the Skills. The User would have to switch the toggle if he doesn't want to include the Skills.

Definition of Done:
- The User can now chose whether to include the "Kernkompetenzen - Skills" in the exported CV or only the "Notizen" field.
- Test added for that use case

PR #425, opened 2021-03-16
#417: let the user chose if core competences and skills shuld be incl…

rbotteon:
See screenshot:
image


Robin481 (approved):
Sehr cool! Stylischer Toggle :grin:

Issue #391, opened 2020-09-22
Query parameters do break reloading


Robin481:
Currently our application with Keycloak doesn't handle reloading well.

If you're not on a root url reloading the page will present you with a white page and nothing will happen until you go back to the root url. (e.g. you can reload on `/cv_search` but you cannot reload on `/cv_search?q=java`)
This is probably due to the Keycloak parameters being lost or not being reloaded correctly.

Timeboxed: 8h

Definition of Done
- Do some research into what is causing this and how it could be resolved.
- Best case: Reloading via F5 or similar is now possible without breaking the Keycloak flow.

Robin481:
I'm adding another Keycloak Issue as they are probably very closely related.

When a user stays on his profile for a long time e.g. 12 minutes and then wants to export his CV this will give him a 401 unauthorized instead of prompting him to login again.

We will probably have to look at Keycloak as a whole and make changes where necessary to better fit the single-page-application.

Definition of resolved:

- When trying to exporting a CV when the Keycloak session has been invalitated the token-refresh functionality gets executed and the user is still able to export his CV.

m4rc77:
This issue could also be renamed to **Query parameters do break reloading**. The problem can be reproduced pretty easy with the skills page.

Loading of `https://<server>/people/<user-id>/skills` works flawless. 

Trying to open `https://<server>/people/<user-id>/skills?rated=true` results in `Uncaught TypeError: t.default.keycloak is undefined`.

```
18:37:39.051 Uncaught TypeError: t.default.keycloak is undefined
    tokenParsed https://<server>/assets/vendor-4f0e312036b29d1754aa42957d631400.js:13209
    get https://<server>/assets/vendor-4f0e312036b29d1754aa42957d631400.js:1789
    untrack https://<server>/assets/vendor-4f0e312036b29d1754aa42957d631400.js:4337
    get https://<server>/assets/vendor-4f0e312036b29d1754aa42957d631400.js:1789
    r https://<server>/assets/vendor-4f0e312036b29d1754aa42957d631400.js:1747
    Oe https://<server>/assets/vendor-4f0e312036b29d1754aa42957d631400.js:1764
    we https://<server>/assets/vendor-4f0e312036b29d1754aa42957d631400.js:1763
    get https://<server>/assets/vendor-4f0e312036b29d1754aa42957d631400.js:2463
    value https://<server>/assets/frontend-6cdfe8f45c85d6c57e3c1472ec0cf54b.js:1760
[...]

PR #429, opened 2021-03-22
Feature 53173 keyclaok docker compose

iimsand:
Run skills in docker with keycloak.

There is a problem with the logout button. If you press logout, the message `error: "No JWT token provided"` is shown up and you have to change the url to http://localhost:8080 for a new clean login. The forwarding is not working correct because the Referrer Policy is "no-referrer-when-downgrade". The problem may also have a connection with #391.


Robin481 (approved):
Merge es mal so da es schon um einiges besser ist.
Wegen der Konsistenz in verschiedenen Browsern habe ich ein Follow-Up Ticket gemacht. (https://github.com/puzzle/skills/issues/439)

Issue #383, opened 2020-06-26
Clear Javascript search query after cv_search (3)


Robin481:
How to reproduce:
- In CV_Search search for a Term and click on the found profile and let yourself be redirected to where your term was found. 
- Go somewhere else and then get back to the Profile you've just been on
- You will still jump to the search query that you used in the cv_search.

Please clear the search query from wherever it gets saved to.

Definition of Done:
- Clicking on a profile that you previously found a cv_search result in should not jump you to that place anymore when you visit said profile again. 
- The jump-to function should only be executed when coming from cv-search

PR #431, opened 2021-03-22
#383: remove set value of existing search query

rbotteon:


Robin481 (changes requested):
Heya 
The "problem" described in https://github.com/puzzle/skills/issues/383 still occurs in the same way. 
(@rbotteon I'm aware you're not responsible for this anymore. I'll pass it on to someone else.)

PR #440, opened 2021-04-19
search query is now  reset

marc89763:
but one frontend test fails

### DIFF
--- a/frontend/app/controllers/cv-search.js
+++ b/frontend/app/controllers/cv-search.js
@@ -1,0 +1,6 @@
+import Controller from "@ember/controller";
+
+export default class CvSearchController extends Controller {
+  queryParams = ["q"];
+  q = null;
+}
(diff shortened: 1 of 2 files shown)


marc89763:
Test "redirect to last skill page using same rated filter" now fails because the **personId** and **rated** query parameters are reset as well. I don't understand why. Used this  [approach](https://medium.com/trouva-product-blog/resetting-query-parameters-when-entering-a-route-in-ember-js-afa8555a1659). Also tried it with Route.resetController as proposed [here](https://guides.emberjs.com/v2.14.0/routing/query-params/#toc_sticky-query-param-values).

marc89763:
Fixed bug.
**Before:** In CV-Search after clicking on gefunden in e.g. _Kompetenzen_ and then returning to CV-Search, the search results were still shown but without the search query in  searchbar and URL.
**After:** In CV-Search after clicking on found in e.g. _Kompetenzen_ and then returning to CV-Search, the search results are no longer shown also without search query in searchbar and URL. When searching for sth. (and not clicking on gefunden in e.g. Kompetenzen) and navigating somewhere else and then returning to CV-Search, the search results are still there as is the search query in searchbar and URL.

Robin481 (approved):
LGTM!

PR #441, opened 2021-04-19
Feature/354 skillset read only for non admin

EmreAlma:


Robin481 (changes requested):
Hey @EmreAlma 
Please add translations for the "You are not allowed to edit" notifications.
Also the Issue asks for tests so please add some tests that for example test that a "non-Admin" user is not able to edit the skills-list.

EmreAlma:
Hi @Robin481, translations for the "You are not allowed to edit" notifications added, and I wrote some easy tests. Would you have a look, please ?

PR #442, opened 2021-04-20
Feature/378 CV search jump to frontend tests

wmvong:
All tests pass when using `ember test --module="Acceptance | person jump to" ` but fail using `rake spec:frontend`.
The message I get when they fail is: **Element not found when calling `click('.cv-search-found-in-link')`**, so I assume the selector might be wrong.


Robin481:
Heya @wmvong 
Just had a quick look at this but couldn't delve too deep into it.
A few pointers that might help you for now: 
- Don't hardcode the ID's of the expected URL because on different environments the fake profiles will not have the same id. Access the ID of the profile directly instead of hardcoding it, similarly to this: https://github.com/puzzle/skills/blob/25d9ebb68dee08f6b284a922d1dfadc7d6b7e4c5/frontend/tests/acceptance/skill-search-test.js#L54
- I got the errors that led me to above conclusion by running `yarn test` (meaning it got further than the place where your problem occurs) as `yarn test` uses `ember test` you would probably need to figure out what differences there are between running your tests in isolation as opposed to the full suite.

If this doesn't help we can try to figure something out on Monday. 
Also maybe @TheWalkingLeek has some inputs here?

Robin481:
Hey @wmvong 
Just checked your last commit. (I get notifications every time any of you pushes something 😁 )  
I think you might have misunderstood me.
Here: https://github.com/puzzle/skills/pull/442/commits/7956312a14dfb8e4386c73c9ff0336c14e5eb23c#diff-b4c313fe499db2fdb6d6bb5cc44dce0de8b29014bc2d5f789b284b46cdcf1713R28
it doesn't make sense to get the ID of Junit. You need to get the ID of the person that you want to visit.

The example I gave above is of a Skill test where we test that it correctly redirects to the intended Skill.
In your case you are testing whether it correctly redirects to a person, therefore you need to get the ID of said person.

Robin481:
Hey @wmvong 
First of all: The tests look really nice, good job!

However, I think we can extend on them a bit as for now they all check very similar things.

I suggest the two following extensions:
- Make sure that one of the tests has multiple results in the CV-Search and then select the 2nd Person that shows up as a result. E.g. You search for "time travel" and you get two people as a result let's say "Ben" and "Ken" (creative I know). You then need to figure out how to press the 2nd of the links accessed by `document.querySelector(".cv-search-found-in-link");` (hint: https://stackoverflow.com/questions/22902447/how-to-get-the-second-match-with-queryselector)
Therefore this test will make sure that if there are multiple results that it doesn't jump to the wrong one for whatever reason.

- What you don't test is whether the page actually scrolls down to the result found in the profiles. You test the URL which works perfectly fine. I suggest in one of the tests you add something that tests the Window Scroll state. 
For example, you search for "Rocket science" which will yield you a result in one of the test user's projects. You then click the button and check the URL as you do  in the others test but now also additionally check whether the page actually has scrolled down. (Maybe you can just check something like that: https://stackoverflow.com/questions/4096863/how-to-get-and-set-the-current-web-page-scroll-position) 

I think those two additions would make the tests more thorough and actually test 3 different things instead of all doing basically the same thing. I know this hasn't been specifically discussed in the original Issue but I think it still fits in the context of writing proper tests for this. As we only planned 2 Sprint Points for this review I would suggest that you don't spend more than 5 hours on this (so basically timeboxed) and if you don't manage to do it in that timeframe we just leave the tests as they are currently. Hit me up if there are any questions :v:

wmvong:
Hiya @Robin481
I managed to do the first extension, but got stuck on the second. Checking the window scroll state always returns 0. I've tried out different expressions but nothing seems to work. :/ I've reached the 5 hours limit so I'll leave it at that now?

Robin481:
Heya @wmvong 
Yea just leave it at the 5 hours 👍 
Would be lovely if you could create a follow-up Issue so we have this documented.

PR #448, opened 2021-05-08
Feature/427 skill search interest filter

marc89763:
Works as intended, but probably isn't perfect yet.


Robin481 (approved):
You know what, that works perfect and looks really good, good job!
The only thing I was considering was doing some refactoring on this code: https://github.com/puzzle/skills/pull/448/files#diff-ca1b69be101a7b9d931a7f8e1c5381db8f30d344c269f833af9aec1574370e73R30 but I think readability is more important in this case so let's just leave it!

You can merge :partying_face:

PR #490, opened 2023-02-13
Feature/482/rails upgrade

lkleisa:


Robin481 (changes requested):
Hello
Code changes look reasonable and the server still runs, so far so good.
However, an important part of upgrading the application is changing the framework defaults to the new version https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#configure-framework-defaults
Please do that and then we can merge this :)

config/application.rb
```diff
@@ -28,6 +28,9 @@ class Application < Rails::Application
     config.autoload_paths += %W( #{config.root}/app/uploaders) #
     config.i18n.default_locale = :de
 
+    # Bullet tries to add finish_at to statement, which does not exist anymore
```
Robin481, line 31: to **insert** statement

Robin481 (changes requested):

Robin481 (approved):

PR #494, opened 2023-02-13
implement branch adress seeder

njaeggi:


db/seeds/support/branch_adress_seeder.rb
```diff
@@ -0,0 +1,18 @@
...
+    branch_short_names.each_with_index do |branch_short_name, index|
+      seed_branch_adress(branch_short_name, branch_adress_informations[index.to_i])
+    end
+  end
+
+  private
+
+  def seed_branch_adress(branch_short_name, adress_information)
+    BranchAdress.seed_once(:adress_information) do |b|
+      b.short_name = branch_short_name.to_s
```
Robin481, line 13: You already pass strings therefore `to_s` is not needed and basically does nothing.
Next time go through your methods with `pry` if you want to check that.

db/seeds/support/branch_adress_seeder.rb
```diff
@@ -0,0 +1,18 @@
+# encoding: utf-8
+class BranchAdressSeeder
+  def seed_branch_adresses(branch_short_names, branch_adress_informations)
+    branch_short_names.each_with_index do |branch_short_name, index|
+      seed_branch_adress(branch_short_name, branch_adress_informations[index.to_i])
```
Robin481, line 5: index is already an integer therefore `to_i` is not needed and does nothing.

Robin481 (changes requested):
Remove the unneeded methods and then we can merge

Robin481 (approved):

Issue #354, opened 2020-03-27
Skillset Read-only for non-admin


Robin481:
Currently only Admins can view the Skillset page. 

It should become read-only for all non-admin members. Meaning these non-admin members will be able to view the page but won't be able to edit anything.
For non-admin members there should be a little warning message explaining to them that they aren't able to edit because they are not admin. This should also just be visually implied. (e.g. Buttons greyed out)

**Definition of read-only**
The only thing non-admin users aren't allowed to do is press the edit button (next to the skills) and the new-skill button. All the other functions are allowed (e.g. search and the skills overlay that show all the people that have said skill).

**Be aware**
There are verifications for the admin role happening in the front as well as the back end. Make sure that both ends are covered and can't be accessed through some workaround. 

Definition of Done:
- The Skillset page is now read-only but accessible for non-admin users.
- A little warning message explains the non-admin users that they can't edit. 
- Nothing changes for admin users
- Tests :hatched_chick:

Robin481:
For https://github.com/puzzle/skills/issues/308 we had to allow the users to create skills, for that we just skipped the create call from authorization: https://github.com/puzzle/skills/commit/baf0cd6ff703aeb042d60af26bf99eb25811d144

This needs to be then done smarter here.

Robin481:
https://github.com/puzzle/skills/pull/441

Feature has been implemented, needs to be reviewed and merged. Might need small updates if we update some part of the application before that.

lkleisa:
Is done here: https://github.com/puzzle/skills/pull/495

PR #495, opened 2023-02-15
Allow non admin people to access skillset

lkleisa:
[Issue 354](https://github.com/puzzle/skills/issues/354)
The skillset page should become read-only for all non-admin members. Meaning these non-admin members will be able to view the page but won't be able to edit anything.

### DIFF
--- a/frontend/app/templates/components/skills-list-row-show.hbs
+++ b/frontend/app/templates/components/skills-list-row-show.hbs
@@ -29,7 +29,7 @@
   {{if this.skill.portfolio this.skill.portfolio "-"}}
 </td>
 <td>
-  <a role="button" onclick={{this.startEditing}} class="edit-buttons skill-edit-button" style='{{unless this.isAdmin "-webkit-filter: grayscale(100%);"}}'>
+  <a role="button" onclick={{this.startEditing}} class="edit-buttons skill-edit-button {{unless this.isAdmin "grayed-out"}}">
     <i class="fa fa-edit"></i>
   </a>
 </td>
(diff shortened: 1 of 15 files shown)


lkleisa:
I could copy mostly the code from [Old PR 441](https://github.com/puzzle/skills/pull/441). I added grayed out buttons: 
Non admin members have now the function to view the skillset page. They can view all the things. The create new skill and the edit button are grayed out for non admins and on click, there is an error message in a toastr pop up.  
Tests in frontend are adjusted. Currently, they are failing, because I added keyCloakService to frontend/app/components/skills-list-row-show.js. This service must be injected in the Test frontend/tests/integration/components/skills-list-row-show-test.js. I couldn't do it and had no help. 
Maybe, changes in backend are missing. In requirements, it says, that we have to make sure that both ends are covered and can't be accessed through some workaround.

Robin481:
Looks good so far, goob job! We'll look at it together tomorrow :smile_cat:

frontend/app/templates/components/skills-list-row-show.hbs
```diff
@@ -29,7 +29,7 @@
   {{if this.skill.portfolio this.skill.portfolio "-"}}
 </td>
 <td>
-  <a role="button" onclick={{this.startEditing}} class="edit-buttons skill-edit-button">
+  <a role="button" onclick={{this.startEditing}} class="edit-buttons skill-edit-button" style='{{unless this.isAdmin "-webkit-filter: grayscale(100%);"}}'>
```
Robin481, line 32: it probably makes more sense to add the css class conditionally than putting the style here. Furthermore I thinkg webkit filter is Firefox specific if i am not mistake, make sure to check that.

Robin481 (approved):
Nice :partying_face:

PR #502, opened 2023-02-27
Improve frontend list components

njaeggi:


Robin481 (changes requested):
Looking quite nice.
Besides the code feedback one thing:
`save-and-new` works when creating a new row but never when editing an already existing one. Fix that and my code feedback and we're good :partying_face:

Robin481 (approved):
Good Job :partying_face: :balloon:

PR #519, opened 2023-03-08
Feature/351 refactor person component with BaseFormComponent

kcinay055679:


frontend/app/components/base-form-component.js
```diff
@@ -32,6 +32,7 @@ export default class BaseFormComponent extends Component {
       .then(savedRecords => {
         this.handleSubmitSuccessful(savedRecords);
         this.afterSubmit();
+        return true;
```
njaeggi, line 35: why was this added?, just curious

frontend/app/components/person-form.js
```diff
@@ -0,0 +1,165 @@
...
+    let create = this.record.id;
+    super.submit(person).then(
+      r =>
+        r &&
+        super
+          .submit(attributes)
+          .then(r => r && this.savePicture(picturePath))
+          .then(r => r && this.afterSuccess(create))
+    );
+  }
```
njaeggi, line 36: what does prettier say here :grin:

frontend/app/components/person-form.js
```diff
@@ -0,0 +1,165 @@
...
+    personRole.personRoleLevel = personRoleLevel;
+  }
+
+  @action
+  setNationality(country) {
+    this.record.nationality = country[0];
+  }
+
+  @action
+  setNationality2(country) {
```
njaeggi, line 126: don't use magic numbers, maybe name this action something like "setSecondNationality"

frontend/app/components/picture-edit.js
```diff
@@ -1,67 +1,36 @@
...
+  constructor() {
+    super(...arguments);
+    this.picturePath = this.args.imagePath;
+  }
+  @action
+  uploadImage(e) {
+    let file = e.target.files[0];
     if (!/\.(?:jpe?g|png|gif|svg|bmp)$/i.test(file.name)) {
-      this.get("notify").alert("Invalider Datentyp");
+      this.notify.alert("Invalider Datentyp");
```
njaeggi, line 19: add translation

frontend/app/components/picture-edit.js
```diff
@@ -1,67 +1,36 @@
...
       return;
     }
 
     if (file.size > 10000000) {
       //10MB
-      this.get("notify").alert("Datei ist zu gross, max 10MB");
+      this.notify.alert("Datei ist zu gross, max 10MB");
```
njaeggi, line 25: add translation

frontend/app/components/picture-edit.js
```diff
@@ -1,67 +1,36 @@
...
-
-  didInsertElement() {
-    this.$(".img-input").on("change", e => {
-      if (e.target.files.length) {
-        this.uploadImage(e.target.files[0]);
-        e.target.value = null;
-      }
-    });
+    this.picturePath = URL.createObjectURL(file);
+    this.args.person.picturePath = URL.createObjectURL(file);
```
njaeggi, line 29: nice work! :rocket:

frontend/app/models/person.js
```diff
@@ -6,7 +6,7 @@ const Person = DS.Model.extend({
   birthdate: DS.attr("date"),
   picturePath: DS.attr("string"),
   location: DS.attr("string"),
-  maritalStatus: DS.attr("string"),
+  maritalStatus: DS.attr("string", { defaultValue: "single" }),
```
njaeggi, line 9: is this really needed?, since it wasn't needed before, or was this a bug?

frontend/tests/acceptance/create-person-test.js
```diff
@@ -2,14 +2,20 @@ import { module, test, skip } from "qunit";
...
+import { click, currentURL } from "@ember/test-helpers";
 import setupApplicationTest from "frontend/tests/helpers/setup-application-test";
-import { currentURL } from "@ember/test-helpers";
 import { selectChoose } from "ember-power-select/test-support";
+import Notify from "ember-notify";
 
 module("Acceptance | create person", function(hooks) {
   setupApplicationTest(hooks);
 
+  const notifyStub = Notify.extend({
```
njaeggi, line 13: really nice! :1st_place_medal:

frontend/tests/acceptance/edit-person-test.js
```diff
@@ -11,7 +11,7 @@ import { fillIn } from "@ember/test-helpers";
 module("Acceptance | edit person", function(hooks) {
   setupApplicationTest(hooks);
 
-  test("/people/:id edit person data", async function(assert) {
+  skip("/people/:id edit person data", async function(assert) {
```
njaeggi, line 14: why do we skip this now, is this because of the pikadate bug?

njaeggi (dismissed):

frontend/app/components/person-form.js
kcinay055679: That wasn't me who introduced this name

frontend/app/components/person-form.js
njaeggi: I would add another line between the function, seems weird that prettier says this is fine :thinking:

frontend/app/models/person.js
kcinay055679: This is necessary now

frontend/app/components/person-form.js
kcinay055679: Prettier wants it exactly like this

frontend/tests/acceptance/edit-person-test.js
Robin481: Then add a comment why it is skipped :wink:

frontend/app/models/person.js
Robin481: Maybe elaborate a bit don't just say yes? :grin:

kcinay055679: It's easier and more readable to set the maritalStaus default value right in the model itself.
Is was defined in the logic of the selector previously.

kcinay055679:
After the rebase on master two tests failed, i tried to fix them but due to a bug in the pikaday addon i could not solve the bug, or at least i think thats the source of the error. Then i tried to simply skip those tests but then ember hit with me with the exception that i have to call setupApplicationContext before i can call visit(). But first i don't call visit() anywhere and second, even if i called setupApplicationContext the errors still appears.

frontend/config/environment.js
```diff
@@ -54,7 +54,7 @@ module.exports = function(environment) {
   if (environment === "test") {
     // Testem prefers this...
     ENV.locationType = "none";
-
+    ENV.APP.autoboot = false;
```
Robin481, line 57: Is this really needed? I am assuming it fixed your `same root element` error?
what happens if you stop splitting the person-form test into 2 different modules?

frontend/tests/acceptance/create-person-test.js
```diff
@@ -0,0 +1,157 @@
...
+module("Acceptance | create person", function(hooks) {
+  setupApplicationTest(hooks);
+
+  const notifyStub = Notify.extend({
+    alert(message, options) {
+      options.closeAfter = null;
+      return this.show("alert", message, options);
+    }
+  });
+  /// Skip this test since there is a bug wtih the @renderInPlace attribute  of the powerselect
```
Robin481, line 19: Did you create some sort of follow-up for this?

frontend/tests/acceptance/create-person-test.js
```diff
@@ -0,0 +1,157 @@
...
+      "Email muss ausgefüllt werden"
+    );
+    assert.equal(
+      document.querySelectorAll(".ember-notify")[1].querySelector(".message")
+        .innerText,
+      "Email Format nicht gültig"
+    );
+  });
+
+  // Skip this test since there is a bug wtih the @renderInPlace attribute  of the powerselect
```
Robin481, line 125: Same here. follow up?

frontend/tests/pages/people-new.js
```diff
@@ -12,16 +12,17 @@ const { resolve } = RSVP;
...
     toggleNationalitiesCheckbox: clickable("#toggle-nationalities-id"),
+    newRoleButton: clickable("[data-test-new-function]"),
 
     toggleNewForm() {
       return this.newForm;
     },
 
     async createPerson(person) {
-      await Object.keys(person).reduce(
+      Object.keys(person).reduce(
```
Robin481, line 25: Does this not cause a problem if it isn't `await` anymore?

Robin481 (changes requested):
Nice work. A few notes and this:

- The translations for `person-role` errors are missing (create an empty role and press save).
- The errors for `person-role` only show up if it is the only error (fill in everything but Wohnsitz and create an empty Role -> press Save -> it will only show the Wohnsitz error -> Fill in a Wohnsitz and only then will it show the `person-role` errors).
- Translations for Language Skills are missing as well and the error above also occurs for the languages (create a new empty language skill and press save). 
- When creating a profile from another profile view and pressing save you will be redirected to the previously visited profile instead of the newly created one. (go to /people/1 -> New profile -> Save -> you are back to /people/1 instead of /people/newProfileId

frontend/tests/acceptance/create-person-test.js
kcinay055679: https://github.com/orgs/puzzle/projects/5/views/1?pane=issue&itemId=27710056

frontend/tests/acceptance/create-person-test.js
kcinay055679: https://github.com/orgs/puzzle/projects/5/views/1?pane=issue&itemId=27710056

Robin481 (changes requested):
Hello.
Looking way better, but I figured the error out that I wasn't able to reproduce when we checked together.

New Profile -> Add a picture -> Fill out everything -> Add a role and press save -> you will get the role-errors -> Fix them and save -> The User will be created with the default picture instead of the one uploaded :grin:

Robin481 (approved):
:rocket:

PR #539, opened 2023-04-20
feature/516-cv-search-go-back-with-cookie

lkleisa:
CV Search go back with cookie, new pr from https://github.com/puzzle/skills/pull/533

### DIFF
--- a/frontend/app/helpers/get-cookie.js
+++ b/frontend/app/helpers/get-cookie.js
@@ -1,0 +1,10 @@
+import { helper } from "@ember/component/helper";
+
+export function getCookie() {
+  let cookieArray = document.cookie.valueOf().split(";");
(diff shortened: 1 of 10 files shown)


frontend/app/helpers/get-cookie.js
```diff
@@ -0,0 +1,10 @@
+import { helper } from "@ember/component/helper";
+
+export function getCookie() {
+  let cookieArray = document.cookie.valueOf().split(";");
+  let cookie = cookieArray[cookieArray.length - 1];
```
Robin481, line 5: I think we should probably add a regex here to make sure we get the correct cookie instead of getting it based on the position which is definetly not fixed.

Robin481 (changes requested):

Robin481 (approved):

PR #592, opened 2024-02-05
Update Workflows

ioboi:
With this PR, the `SBOM` workflow uses the latest tooling and works again.
I also refactored the `Rails` workflow to make it easier to extend. The tests now use PostgreSQL 16.


Robin481 (approved):
:kissing_heart:

PR #597, opened 2024-02-06
Recreate skills list 

kcinay055679:


Robin481 (changes requested):
Good start.
A few things to improve:

- No need for enumeration, just remove the first row.
- Align the table headers with the contained data.
- As the ticket stated: "Make it visually appealing and as close as possible to the old Ember View" please make the table visually more close to the old one. (grey-white switching, lines between the rows, proper header differentiation) I know it said `Bonus` but this should be easy enough. 

And another thought:
I know we sort of already discussed this but I don't think we should use the whole screen width. At least not in this view here. The introduced spacing makes it only more complicated to follow the data throughout the rows. 
As this is a very general thing for the whole application please add it somewhere in the base layout so that it doesn't use the whole width throughout the application.

Robin481 (approved):

PR #610, opened 2024-02-20
Feature/601 people dropdown

kcinay055679:


.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict = true
```
Robin481, line 1: Needed?

Gemfile
```diff
@@ -57,7 +57,7 @@ group :development, :test do
 end
 
 group :development do
-  gem 'bullet'
+  gem 'bullet', '~> 7.1', '>= 7.1.6'
```
Robin481, line 60: Why the fixed version?

Gemfile
```diff
@@ -71,7 +71,7 @@ group :test do
   gem 'capybara'
   gem 'selenium-webdriver'
   gem 'simplecov'
-  gem 'webdrivers'
+  gem 'webdrivers', '~> 5.3', '>= 5.3.1'
```
Robin481, line 74: Why the fixed version?

Procfile.assets
```diff
@@ -0,0 +1,2 @@
+js: yarn build --watch
+css: yarn watch:css
```
Robin481, line 2: nice!

app/controllers/people_controller.rb
```diff
@@ -2,17 +2,6 @@
...
-  self.permitted_attrs = %i[birthdate location
-                            marital_status updated_by name nationality nationality2 title
-                            competence_notes company_id email department_id shortname]
-
-  self.nested_models = %i[advanced_trainings activities projects
-                          educations language_skills person_roles
-                          people_skills categories]
-
-  self.permitted_relationships = %i[person_roles people_skills]
-
```
Robin481, line 15: I would not completely delete this but rather just comment it out as we will probably use this later anyways.

app/javascript/controllers/dropdown_controller.js
```diff
@@ -0,0 +1,11 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="dropdown-links"
+export default class extends Controller {
+  connect() {
+  }
+
```
Robin481, line 7: Remove linebreak

app/views/people/_search.html.haml
```diff
@@ -0,0 +1,3 @@
+= form_with do |f|
+  %section{"data-controller"=>"dropdown"}
+    = f.collection_select :person_id, Person.all.sort_by(&:name), :id, :name, {:selected => person.id} , {class: "form-select w-100", "data-action": "change->dropdown#handleChange", "data-value": "/people/"}
```
Robin481, line 3: Handle case when there are no people in the database please :) 
(safe operator & olé)

Robin481 (changes requested):
Looking very good!
Some nitpicking and then we are good :rocket:

.npmrc
kcinay055679: I added this to enforce yarn and restrict the use of npm

Gemfile
kcinay055679: Otherwise an error occurs due to a compatibility error between chrome and webdriver

Gemfile
kcinay055679: Oops

app/views/people/_search.html.haml
kcinay055679: Person.all.sort_by(&:name) already returns an empty array (instead of an exception) if no people found, do you mean smth other?

Robin481 (approved):

PR #611, opened 2024-02-23
TESTING Feature/590 add skill 

kcinay055679:


app/controllers/skills_controller.rb
```diff
@@ -1,12 +1,9 @@
 # frozen_string_literal: true
 
-require 'keycloak_tools'
-
 class SkillsController < CrudController
   include ExportController
-  include KeycloakTools
 
-  before_action :authorize_admin, except: %i[index show create unrated_by_person]
+  # before_action :authorize_admin, except: %i[index show create unrated_by_person]
```
Robin481, line 6: delete

spec/features/skills_form_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    @portfolio = "passiv"
+  end
+
+  it 'creates the skill when the form is submitted' do
+    fill_in 'skill_title', with: 'New Skill Title'
+    select @category.title, from: 'skill_category_id'
+    select @radar, from: 'skill_radar'
+    select @portfolio, from: 'skill_portfolio'
+    click_button 'Skill erstellen'
+    all 'tbody tr', count: 6
```
Robin481, line 19: Seems like it is a little bit too unspecific. If we added a table somewhere else this test would pass for the wrong reason.
Maybe check for the 6 names of the skills instead?

spec/features/skills_form_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+
+describe 'Skill Form', type: :system, js:true do
+  before do
+    visit skills_path
+    click_link(href: new_skill_path)
```
Robin481, line 7: Why do you have to set the href?

spec/controllers/skills_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+        expect(response).to render_template("index")
+      end
+
+      it 'post returns all skills ' do
+        title = 'new skill'
+        category_id = categories(:java).id
+        radar = "hold"
+        portfolio = "passiv"
+
+        post :create , params: { skill: { title: title, category_id: category_id, radar: radar, portfolio: portfolio} }
```
Robin481, line 18: Just pass the values straight into the creation call here instead of defining them on lines 13-16.

Robin481 (changes requested):
Looking good.
Small adjustments and then merge time :rocket:

spec/features/skills_form_spec.rb
kcinay055679: Cause it's likely we change the name of the button but the href stays consistent

spec/controllers/skills_controller_spec.rb
kcinay055679: but if i do it like this i can use the same vars to check if the value is actually the same

Robin481 (approved):
:rocket:

PR #612, opened 2024-02-23
add auth gem

kcinay055679:


Gemfile
```diff
@@ -14,6 +14,7 @@ gem 'config'
 gem 'countries'
 gem 'cssbundling-rails'
 gem 'database_cleaner'
+gem 'devise', '~> 4.9'
```
Robin481, line 17: Do we need to specify a version range or can we just leave it empty as 4.9.3 is the newest one anyway?

app/controllers/application_controller.rb
```diff
@@ -1,4 +1,11 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
+  before_action :authenticate_person!
+
+  def check_admin
+    return false if helpers.admin?
```
Robin481, line 7: Reverse logic?

app/models/person.rb
```diff
@@ -92,4 +94,33 @@ def picture_size
 
     errors.add(:picture, 'grösse kann maximal 10MB sein')
   end
+
+  class << self
+    def from_omniauth(auth) # rubocop:disable Metrics/AbcSize
+      person = where(email: auth.info.email).first_or_create do |user|
+        user.name = auth.info.name
+        user.birthdate = DateTime.new(2020, 1, 1)
```
Robin481, line 102: We have to make many assumptions here which isn't nice.
Do we need to create a Person for the login to make sense?

app/models/person.rb
```diff
@@ -92,4 +94,33 @@ def picture_size
...
+        user.company = Company.first
+      end
+      set_admin(person, auth)
+    end
+
+    private
+
+    def set_admin(person, auth)
+      person.is_admin = admin?(auth)
+      person.save
```
Robin481, line 116: What if we remove the admin flag from a person in the keycloak interface?
What if the person logs in and that person already has a profile? (will be the case on PROD)

lib/auth_config.rb
```diff
@@ -0,0 +1,52 @@
...
+    def admin_role
+      get_var_from_environment(:admin_role, required: false)
+    end
+
+    private
+
+    def get_var_from_environment(key, required: true)
+      if !locale? && required
+        ENV[key.to_s] || raise("Environment variable not set: '#{key}'")
+      elsif !locale?
```
Robin481, line 33: `unless`

lib/auth_config.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def admin_role
+      get_var_from_environment(:admin_role, required: false)
+    end
+
+    private
+
+    def get_var_from_environment(key, required: true)
+      if !locale? && required
```
Robin481, line 31: `unless`

Robin481 (dismissed):

lib/auth_config.rb
kcinay055679: If i use unless then I'm not able to use elseif. 
Additionally there isn't a huge diffrence beween 
`if !locale? && required` and `unless locale? && !required` since in both cases there is one negated param

lib/auth_config.rb
kcinay055679: look up reason above

PR #624, opened 2024-03-13
Feature/619 show people modal

MarcEgliP:


app/views/skills/show.html.haml
```diff
@@ -1,7 +1,32 @@
...
+    - @skill.people_skills.each do |person_skill|
+      - if person_skill.level
+        %div.d-flex.flex-row.align-items-center.skill-row
+          %div.col-2
+            = link_to Person.find(person_skill.person_id).name, person_path(person_skill.person_id)
+          %div.col-2.d-flex.justify-content-center
+            %p.m-0= @skill.title
+          %div.col-2
+            %div.d-flex.flex-column
+              %label.form-label.text-gray= t ExpertiseTopicSkillValue.skill_levels.key(person_skill.level - 1)
```
Robin481, line 12: General thing for this whole file:
From what we learned above `level: 1` and `interest: 1` are the lowest possible value that can be selected. Therefore a people_skill with  `level: 1` and `interest: 1` should display both bubbles completely to the left which currently is not the case.

app/views/skills/show.html.haml
```diff
@@ -1,7 +1,32 @@
...
+              %input.form-range.w-75.level{type: "range", min: 0, max: 5, value: person_skill.level, tabindex: -1}
```
Robin481, line 13: What does `tabindex: -1` accomplish here?

spec/controllers/people_controller_spec.rb
```diff
@@ -11,15 +11,15 @@
     it 'should save nationality2 as nil if checkbox is unchecked and save it if it is checked' do
       expect(person["nationality2"]).not_to be(nil)
       put :update, params: {id: person["id"], person: person, has_nationality2: {checked: "0"}}
-      response.code.should == "302"
-      response.should redirect_to(person_path(person["id"]))
+      expect(response.code).to include("302")
```
Robin481, line 14: Hmm I think the way it was in the first place was correct. We expect response code `302` nothing else. Although not realistic in this case `3020` would also match

spec/fixtures/people_skills.yml
```diff
@@ -35,6 +35,12 @@ alice_rails:
   certificate: false
   core_competence: false
 
+alice_bash:
+  skill: bash
+  person: alice
+  certificate: false
```
Robin481, line 41: Set this to `0` instead of nil

Robin481 (changes requested):
Some small things, after that we are ready for liftoff :rocket: :rocket: :rocket: :rocket: :rocket:

Robin481 (approved):
:cheese:

PR #629, opened 2024-03-19
Feature/603 notes core competences

MarcEgliP:


Robin481 (changes requested):
Looking good.

Please change the 2 small things we've discussed and then we are good to go:
- Give the text box more width
- Seperate core_competences that are in the same category with a small green bubble instead of giving them a new line

Robin481 (approved):
:rock:

PR #651, opened 2024-04-09
Re-Add Skill - People Search

MarcEgliP:


Robin481 (changes requested):
First of all a few visual feedbacks before I will review the code:
- Sort the skills in the dropdown by alphabet
- Add a link to the user's profile to the username
- Seperate the search results by some kind of bounding line
- Add some kind of bounding line to the sides as well (for this and the last one orientate yourself on the old design)

Robin481 (changes requested):
A few small notes

Robin481 (approved):

PR #657, opened 2024-04-15
Feature/643 multiple people skill search

MarcEgliP:


app/controllers/concerns/param_converters.rb
```diff
@@ -15,4 +15,41 @@ def false?(value)
   def nil_param?(value)
     value == 'null' ? nil : value
   end
+
```
mtnstar, line 18: that's what I meant by creating a class that can handle these search params ... it's not very clean code if those skill search related param helpers are just in this generic concern/module.

so it would be nice to have this in a own class ... something like PeopleSkills::FilterParams class I suggested before.

to me it also looks like there's to much manual work for converting the params ... rails params are pretty standard and I'm sure there's helpers to handle these conversions without to much implementation/code

app/controllers/people_skills/filter_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class PeopleSkills::FilterController < ApplicationController
```
mtnstar, line 3:
```suggestion
class PeopleSkills::FilterFormController < ApplicationController
```

app/controllers/people_skills_controller.rb
```diff
@@ -2,25 +2,42 @@
...
-  def search_level
-    params[:level] ? params[:level].to_i : 1
+  def row_count
+    params[:skill_id].present? ? params[:skill_id].length : 1
   end
 
-  def search_interest
-    params[:interest] ? params[:interest].to_i : 1
+  # rubocop:disable Metrics/AbcSize
+  def extract_query_params
```
mtnstar, line 31: same for this part ... extract the whole logic how to extract these filter params into an own class that is specific for this task

mtnstar:
maybe some inspiration from hitobito: 

- https://github.com/hitobito/hitobito/blob/master/app/views/events/courses/_filter.html.haml#L21
- https://github.com/hitobito/hitobito/blob/master/app/controllers/concerns/events/course_listing.rb#L66
- https://github.com/hitobito/hitobito/tree/master/app/domain/events

app/controllers/people_skills/params.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def skills
+      return [] if @params[:skill_id].nil? || @params[:skill_id].blank?
+
+      @params[:skill_id].select { |skill| skill.present? }
+    end
+
+    def levels
+      @params[:level].presence
```
Robin481, line 17: A method called `levels` should in my world return levels, not check for their presence. How about `levels_present` ?

app/controllers/people_skills/params.rb
```diff
@@ -0,0 +1,70 @@
...
+    def rows
+      @params[:rows].present? ? @params[:rows].to_i : nil
+    end
+
+    def rows_count
+      @params[:skill_id].present? ? @params[:skill_id].length : 1
+    end
+
+    def search_level(row_id)
+      return 1 if levels.nil?
```
Robin481, line 33: `return 1 unless levels`

app/controllers/people_skills/params.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      levels[row_id].present? ? levels[row_id].to_i : 1
+    end
+
+    def search_interest(row_id)
+      return 1 if interests.nil?
```
Robin481, line 39: `return 1 unless levels`

app/controllers/people_skills/params.rb
```diff
@@ -0,0 +1,70 @@
...
+  class Params
+
+    def initialize(params)
+      @params = params
+    end
+
+    def skills
+      return [] if @params[:skill_id].nil? || @params[:skill_id].blank?
+
+      @params[:skill_id].select { |skill| skill.present? }
```
Robin481, line 13: No idea what you are trying to achieve here.
If it is filtering the array for nil values then just do https://apidock.com/ruby/v2_6_3/Array/compact

Robin481: Furthermore as this method doesn't returns skills but skill_ids it should probably be called the part

app/controllers/people_skills/params.rb
```diff
@@ -0,0 +1,70 @@
...
+      levels[row_id].present? ? levels[row_id].to_i : 1
+    end
+
+    def search_interest(row_id)
+      return 1 if interests.nil?
+
+      interests[row_id].present? ? interests[row_id].to_i : 1
+    end
+
+    def search_skill(row_id)
```
Robin481, line 44: better naming

app/domain/people_skills_filter.rb
```diff
@@ -62,6 +58,7 @@ def find_person_skills(entries, person_ids)
   def skills_by_person(entries, person_ids)
     entries.includes(:skill, :person)
            .where(person_id: person_ids, skill_id: skill_ids)
+           .sort_by { |item| item.skill.title }
```
Robin481, line 61: probably add `.downcase`

Robin481 (changes requested):
:rock:

mtnstar (changes requested):
your code / feature has improved a lot since the last review ! :rocket: 

- you extracted the filter params to it's own class as I suggested, but I think the whole code is still a bit hard to understand. Think about better naming and make your code more clean. 
- can we make the filter form url be accessible by turbo only?

config/routes.rb
```diff
@@ -11,14 +11,24 @@
...
 
+  resources :skills do
+    collection do
+      get 'export', to: 'skills#export'
+    end
+  end
+
+  resources :people_skills do
+    collection do
+      get 'filter_form', to: "people_skills/filter_form#index"
```
mtnstar, line 22: can we restrict this to turbo only?

spec/features/people_skill_search_spec.rb
```diff
@@ -2,34 +2,117 @@
 
 describe :people_skills do
   describe 'People Search', type: :feature, js: true do
-    let(:skill) { skills(:rails).attributes }
+    let(:rails) { skills(:rails).attributes }
```
mtnstar, line 5:
```suggestion
    let(:skill) { skills(:rails).attributes }
```

I find it rather confusing if you just read rails further down in the spec ... `rails_skill` would also also be a suitable name maybe

spec/features/people_skill_search_spec.rb
```diff
@@ -2,34 +2,117 @@
...
       expect(page).to have_text('Bob Anderson')
       expect(page).to have_text('Wally Allround')
     end
 
     it 'Should set values of query parameters' do
-      visit(people_skills_path + "?skill_id=#{skill["id"]}&level=3&interest=5")
-      expect(page).to have_select("skill_id", selected: "Rails")
-      expect(page).to have_field("level", with: 3)
-      expect(page).to have_field("interest", with: 5, visible: false)
+      visit(people_skills_path + "?skill_id[]=#{rails["id"]}&level[]=3&interest[0]=5")
```
mtnstar, line 22: this is quite a manual way of providing url params ...
`people_skills_path` could accept arguments as hash and I would prefer it this way since it makes it more human readable

mtnstar (changes requested):
nice nice nice, almost there :rocket: 

try to avoid ignoring any rubocop issues ... rubocop is here to tell you when code is not readable or ugly (to much bricolage)

check other comments

config/routes.rb
MarcEgliP: We are going to restrict the turbo-controllers in another ticket as discussed.

PR #664, opened 2024-04-18
Feature/652-readd-cv-search

RandomTannenbaum:


app/assets/stylesheets/styles.scss
```diff
@@ -92,6 +95,22 @@ pzsh-topbar {
   background-color: #3268a1;
 }
 
+.bg-skills-green {
+  background-color: $skills-green;
+}
+
+.bg-skills-green:hover {
+  background-color: $skills-green-hover;
```
Robin481, line 103: How about a little nesting? 
```
.bg-skills-green {
  background-color: $skills-green;

  &:hover {
    background-color: $skills-green-hover;
  }
}
```

app/assets/stylesheets/styles.scss
```diff
@@ -92,6 +95,22 @@ pzsh-topbar {
...
+}
+
+.bg-skills-search-result-blue {
+  background-color: $skills-search-result-blue;
+}
+
+.bg-skills-search-result-blue:hover {
+  background-color: $skills-green-hover;
```
Robin481, line 111: also nesting please

app/controllers/cv_search_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class CvSearchController < ApplicationController
+
+  def index
+    @cv_search_results = query.nil? || query.length < 3 ? [] : search_results
```
Robin481, line 6: `query.nil? || query.length < 3` I would extrapolate this to a seperate method named something like `should_search` and then use that for the ternary because otherwise the ternary is not readable anymore

app/views/cv_search/index.html.haml
```diff
@@ -0,0 +1,16 @@
...
+  %turbo-frame{id: "search-results"}
+    - if @cv_search_results.blank?
+      Keine Resultate
+    - else
+      - @cv_search_results.each do |result|
+        %div.w-50.d-flex
+          = link_to result[:person][:name], person_path(result[:person][:id]), {class: "bg-skills-green w-50 text-decoration-none text-white ps-1 p-2 rounded-1", "data-turbo": "false"}
+          %div.w-50.d-flex.justify-content-end.align-items-center
+            %div.me-1 gefunden in:
+            = link_to I18n.t("cv_search.#{result[:found_in].split("#")[0].underscore}"), "#{person_path(result[:person][:id])}?q=#{params[:q]}", {class: "bg-skills-search-result-blue w-50 text-decoration-none text-white ps-1 p-2 rounded-1 text-center", "data-turbo": "false"}
```
Robin481, line 15: `cv_search.#{result[:found_in].split("#")[0].underscore}` i think this and 
`#{person_path(result[:person][:id])}?q=#{params[:q]}` this deserve their own methods as to keep the view simpler.

app/javascript/controllers/search_controller.js
```diff
@@ -0,0 +1,12 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+    timeout;
+    submitWithTimeout(e) {
+        const form = e.target.parentElement;
+        clearTimeout(this.timeout);
+        this.timeout = setTimeout(() => {
+            form.requestSubmit();
+        }, 400)
```
Robin481, line 10: The delay makes sense but is probably a bit too long:
According to this: https://stackoverflow.com/questions/4098678/average-inter-keypress-time-when-typing
A user notices a delay starting from 100ms.
Furthermore, the average time between keystrokes is 167ms.
I tried it with 100ms and it just feels the best. 200ms is also fine but the delay is noticeable. 
So let's do 100ms for now and if we ever encounter performance errors we know which number to bump.

Robin481 (changes requested):
Looking really good, good job!

One more thing besides the comments:
When reloading or sending a direct link to http://localhost:3000/cv_search?q=arduino the text will be gone from the input field.
Make sure that if there is a `q` param but no text in the field to populate the field on page load.

Robin481 (approved):

PR #681, opened 2024-04-29
Bug/678 missing translation

MarcEgliP:


app/views/people/index.html.haml
```diff
@@ -1,7 +1,8 @@
 = render partial: "people/search", locals: { person: nil }
 - flash.each do |type, msg|
-  .alert.alert-danger.mt-3
-    = msg
+  - if type == 'error'
```
Robin481, line 3: Well now we don't show any messages besides `error` anymore :thinking: 
Is that really the idea?

Robin481 (changes requested):

PR #685, opened 2024-04-30
Feature/674 limit file-size of avatar

ManuelMoeri:


app/assets/images/huge_picture.jpg
Robin481, line 1: No :joy: 
We can't just use any random picture, does it have some sort of Copyright on it etc. etc. etc.

Instead in the test stub the method that checks for file size and make it return `< 10mb` :smile:

Robin481 (changes requested):

app/assets/images/huge_picture.jpg
ManuelMoeri: Thanks for this crucial input. I fixed it.

Robin481 (approved):

PR #688, opened 2024-05-02
make admin check null save

kcinay055679:


app/models/auth_user.rb
```diff
@@ -23,8 +23,8 @@ def set_admin(person, auth)
     end
 
     def admin?(auth)
-      resources = auth.extra.raw_info.resource_access[AuthConfig.client_id]
-      resources.roles.include? AuthConfig.admin_role
+      resources = (auth.extra&.raw_info&.resource_access & [AuthConfig.client_id]) || nil
```
Robin481, line 26: The safe operator should return nil bi default

Robin481 (changes requested):

PR #703, opened 2024-05-13
Feature/696 show updated at

kcinay055679:


Robin481 (changes requested):
image
Align text to center of dropdown please.

PR #727, opened 2024-06-11
 feature/694 CV-search optionally search skills as well

ManuelMoeri:


app/controllers/api/people/search_controller.rb
```diff
@@ -10,7 +10,7 @@ def index
     private
 
     def search_results
-      PeopleSearch.new(query).entries
+      PeopleSearch.new(query, false).entries
```
Robin481, line 13: I know this is just a fix for code that is probably not used anymore but you could give this variable a name when passing it, just `false` is quite confusing :grin: 

`PeopleSearch.new(query, search_skills: false).entries`

app/domain/people_search.rb
```diff
@@ -35,19 +40,19 @@ def pre_load(people)
...
   def in_associations(person)
-    association_symbols.each do |sym|
-      attribute_name = in_association(person, sym)
-      if attribute_name
-        return format('%<association>s#%<attribute_name>s',
-                      association: sym.to_s, attribute_name: attribute_name)
+    association_symbols.map do |sym|
+      attribute_names = in_association(person, sym)
+      attribute_names.map do |attribute_name|
+        format('%<association>s#%<attribute_name>s',
```
Robin481, line 51: why not just `"#{sym.to_s}\##{attribute_name}"` 
Doesn't this do the same or am I missing something?

Robin481 (changes requested):
Looking good and working nicely.

Could you make sure that "unrated" skills don't show up as a result?

Robin481 (approved):

PR #741, opened 2024-06-27
add new controller and add autoclick

kcinay055679:


Robin481 (changes requested):
Can you not solve this more elegantly instead of clicking the button with Javascript? How did you do it for Cryptopus?

Robin481 (approved):

PR #754, opened 2024-07-29
add delete button

kcinay055679:


Robin481 (changes requested):
Functionality is fine but I don't like the look and naming of the button.

What we did in the old EmberJS Application is hide the "delete" Button in a Dropdown called "Weitere Funktionen" and then call it "Person löschen" in the dropdown.

Furthermore the confirmation message should read "Willst du diese Person wirklich löschen"

Robin481 (approved):

PR #803, opened 2025-02-18
improve skill search feedback

nevio18324:


spec/controllers/people_skills_controller_spec.rb
```diff
@@ -14,10 +14,10 @@
       expect(response.body).to include("Bob Anderson")
     end
 
-    it 'should return no results if skill id is not given' do
+    it 'should return add skill to search if skill id is not given' do
```
ManuelMoeri, line 17: Maybe write 'Should return add skill to search message if skill id is not given'? I think this would be a bit more clear.

spec/features/people_skill_search_spec.rb
```diff
@@ -16,17 +16,17 @@
...
-      visit people_skills_path({skill_id: [skill.id], level: [3], "interest[0]": 5})
+      visit people_skills_path({ skill_id: [skill.id], level: [3], "interest[0]": 5 })
       expect(page).to have_select("skill_id[]", selected: "Rails", visible: false)
       expect(page).to have_field("level[]", with: 3)
       expect(page).to have_field("interest[0]", with: 5, visible: false)
     end
 
-    it 'Should return no results if skill id is not set' do
-      visit people_skills_path({level: [3], "interest[0]": 5})
+    it 'Should return add skill to search if skill id is not set' do
```
ManuelMoeri, line 25: Same thing here.

config/locales/de.yml
```diff
@@ -158,9 +158,22 @@ de:
...
+        '5': Experte
     search:
       found_in: Gefunden in
+      no_match:
+        first_skill: Keine Resultate gefunden, mit dem Skill %{skill} auf dem Level %{level} oder höher und dem Interesse %{interest} oder grösser und
+        last_skill: mit dem Skill %{skill} auf dem Level %{level} oder höher und dem Interesse %{interest} oder grösser.
+        only_skill: Keine Resultate gefunden, mit dem Skill %{skill} auf dem Level %{level} oder höher und dem Interesse %{interest} oder grösser.
+        skill: mit dem Skill %{skill} auf dem Level %{level} oder höher und dem Interesse %{interest} oder grösser und
       no_results: Keine Resultate
+      no_skill: Keine Resultate gefunden, fügen sie einen Skill zur Suche hinzu.
```
ManuelMoeri, line 176: Following thing to think about: Do we ever use 'Sie' or any pronoun in the other messages? If no, we probably should just use 'Keine Resultate gefunden, füge einen Skill zu der Suche hinzu'.

(If you need to change it, do it for every language :1234: )

config/locales/de.yml
```diff
@@ -158,9 +158,22 @@ de:
...
+        '0': Nicht bewertet
+        '1': Azubi
+        '2': Junior
+        '3': Professionell
+        '4': Senior
+        '5': Experte
     search:
       found_in: Gefunden in
+      no_match:
+        first_skill: Keine Resultate gefunden, mit dem Skill %{skill} auf dem Level %{level} oder höher und dem Interesse %{interest} oder grösser und
```
ManuelMoeri, line 171: image

There is a couple of questionable things here. We have some capitalized words like 'Mit' that definitely should not be and the sentence is really, really long. Is there a way to clean this up?

ManuelMoeri (changes requested):

app/views/people_skills/index.html.haml
```diff
@@ -16,7 +16,14 @@
   %turbo-frame{id: "search-results"}
     %div.border.p-3.rounded-bottom.search-results
       - if entries.size == 0
-        = ti "search.no_results"
+        - if filter_params.skill_ids.empty? || filter_params.skill_ids.any? { |x| x.to_i <= 0 }
```
Robin481, line 19: The first part sure, but what does the second part even do?

app/views/people_skills/index.html.haml
```diff
@@ -16,7 +16,14 @@
...
+          = ti "search.no_skill"
+        - else
```
Robin481, line 21: This is all quite overcomplicated and not readable at all.
How about building the whole string in the controller?
So basically all you would have in the view is:
`= search_result_message(filter_params)`

Then in the controller add the above method and build the string dynamically. This allows you to clean up quite a bit of code and split it up into several methods (which you should).

config/locales/de.yml
```diff
@@ -158,9 +158,22 @@ de:
...
+        last_skill: mit dem Skill %{skill} auf dem Level %{level} oder höher und dem Interesse %{interest} oder grösser.
+        only_skill: Keine Resultate gefunden, mit dem Skill %{skill} auf dem Level %{level} oder höher und dem Interesse %{interest} oder grösser.
+        skill: mit dem Skill %{skill} auf dem Level %{level} oder höher und dem Interesse %{interest} oder grösser und
       no_results: Keine Resultate
+      no_skill: Keine Resultate gefunden, fügen Sie einen Skill zur Suche hinzu.
```
Robin481, line 176: I think `füge einen Skill zur Suche hinzu` is also fine, no need to be this formal.

app/controllers/people_skills_controller.rb
```diff
@@ -31,4 +31,20 @@ def extract_required_search_values
     filtered_array = filter_params.skill_ids.zip(filter_params.levels, filter_params.interests)
     filtered_array.reject { |arr| arr.first == '' }.transpose
   end
+
+  def find_skill_title_by_id(id)
```
Robin481, line 35: you can `private` the helper_methods

app/controllers/people_skills_controller.rb
```diff
@@ -31,4 +31,20 @@ def extract_required_search_values
...
+    Skill.find(id).title
+  end
+
+  def get_no_match_translation(index)
```
Robin481, line 39: Try to work on your naming as well.
`get_no_match_translation` doesn't really speak for itself. Especially because it handles several cases at the same time (no skill was found / several skills weren't found) Also the `get` part doesn't help with anything, you call the method so getting is implied. 

Something like `no_results_message` covers them all, says what it actually is and is more readable.

app/controllers/people_skills_controller.rb
```diff
@@ -31,4 +31,20 @@ def extract_required_search_values
...
+    if index.zero? && index == filter_params.rows_count - 1
```
Robin481, line 40: I think you can improve this whole method a lot, this ties in with the outsourcing comment i left in the view. 
Try to split this up into several methods, use variables that speak for themselves etc.

For example `filter_params.rows_count` should be assigned to a variable especially because you use it multiple times but also to make it readable. Something like `skills_searched` speaks volumes.

app/helpers/i18n_helper.rb
```diff
@@ -12,12 +12,16 @@ module I18nHelper
...
   #  - ...
   #  - global.{key}
-  def translate_inheritable(key, variables = {}) # rubocop:disable Metrics/CyclomaticComplexity
+  def translate_inheritable(key, variables = {}) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/AbcSize, Metrics/PerceivedComplexity
     partial = defined?(@virtual_path) ? @virtual_path.gsub(/.*\/_?/, '') : nil
     defaults = inheritable_translation_defaults(key, partial)
     variables[:default] ||= defaults
     variables[:model] ||= model_class&.model_name&.human if respond_to?(:model_class)
-    t(defaults.shift, **variables).upcase_first
+    if !variables.key?(:upcase) || variables[:upcase]
```
Robin481, line 20: Instead of introducing an if else here causing Rubocop to complain about AbcSize and complexity I implore you to try this in one line or split it up into smaller methods. 

Rubocop is here for a reason :wink:

Robin481 (changes requested):
@nevio18324 There you go. 
The feature works quite well but the code needs to be cleaned up quite a bit. :rock:

app/views/people_skills/index.html.haml
Robin481: And why does the test do that? :grimacing:

config/database.yml
```diff
@@ -14,7 +14,7 @@ development:
 
 test:
   <<: *default
-  database: <%= ENV['RAILS_DB_NAME'] || "skills_test" %>
+  database: <%= "skills_test" %>
```
RandomTannenbaum: Revert this

spec/features/people_skill_search_spec.rb
```diff
@@ -16,17 +16,17 @@
...
+      visit people_skills_path({ skill_id: [skill.id], level: [3], "interest[0]": 5 })
       expect(page).to have_select("skill_id[]", selected: "Rails", visible: false)
       expect(page).to have_field("level[]", with: 3)
       expect(page).to have_field("interest[0]", with: 5, visible: false)
     end
 
-    it 'Should return no results if skill id is not set' do
-      visit people_skills_path({level: [3], "interest[0]": 5})
+    it 'Should return add skill to search message if skill id is not given' do
+      visit people_skills_path({ level: [3], "interest[0]": 5 })
```
RandomTannenbaum, line 26: Why set level and interest in this test?

RandomTannenbaum (changes requested):

spec/features/people_skill_search_spec.rb
nevio18324: we test if they are set right later in the test

Robin481 (changes requested):
I am not really happy with the current solution, the sentences are getting way too long.
I think we could probably make it more readable by doing sentences like this:
`No Results found for the following search: Ruby on Rails (2/5), Ruby (2/3), EmberJS (1,5)`

Furthermore your handling of the filter_params can still be improved and made more simple, with the above input try to do that :rocket:

spec/features/people_skill_search_spec.rb
```diff
@@ -42,10 +42,19 @@
...
-      fill_out_row("Bash", 5, 3)
-      expect(page).to have_text("Keine Resultate")
+      skill = "Bash"
+      fill_out_row(skill, 5, 3)
+      expect(page).to have_text("Keine Resultate gefunden mit der folgenden Suche: #{skill} (5/3)")
     end
 
+    it 'Should return no results if no user matches multiple filters' do
+      visit(people_skills_path)
+      skill1 = "Bash"
```
Robin481, line 52: Is that really an improvement over just typing the name twice?

Robin481 (changes requested):
Getting there :crossed_fingers:

spec/controllers/people_skills_controller_spec.rb
```diff
@@ -14,12 +14,6 @@
       expect(response.body).to include("Bob Anderson")
     end
 
-    it 'should return no results if skill id is not given' do
```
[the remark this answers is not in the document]
nevio18324: This is the Test that gave nil as an param which somewhere gets mapped to a 0

spec/features/people_skill_search_spec.rb
```diff
@@ -42,10 +42,17 @@
 
     it 'Should return no results if no user matches filters' do
       visit(people_skills_path)
-      fill_out_row("Bash", 5, 3)
-      expect(page).to have_text("Keine Resultate")
+      skill = "Bash"
```
Robin481, line 45: I assumed you would change this one as well :wink:

spec/features/people_skill_search_spec.rb
```diff
@@ -42,10 +42,17 @@
...
+      fill_out_row(skill, 5, 3)
+      expect(page).to have_text("Keine Resultate gefunden mit der folgenden Suche: #{skill} (5/3)")
     end
 
+    it 'Should return no results if no user matches multiple filters' do
+      visit(people_skills_path)
+      fill_out_row("Bash", 5, 3)
```
Robin481, line 52: I assumed you would change this one as well :wink:

Robin481 (changes requested):
:rocket:

Robin481 (approved):
:book: :bookmark: :bookmark_tabs: :books: :boom:

PR #819, opened 2025-03-03
Make slim select dropdown search non space-sensitive

RandomTannenbaum:


app/domain/skills_filter.rb
```diff
@@ -10,17 +10,23 @@ def initialize(entries, category, title, default_set)
     @default_set = default_set
   end
 
+  # rubocop:disable Metrics/MethodLength
```
ManuelMoeri, line 13: Is this necessary or can this be outsourced properly?

ManuelMoeri (changes requested):
I only found one thing, and it might not even be a thing. Take a look at it. If you can manage to rationally outsource a part of the method and remove the rubocop line, that would be great.

Robin481 (approved):
:rock: :jack_o_lantern:

PR #820, opened 2025-03-03
some dates in cv missing when downloading anonymous

nevio18324:


nevio18324:
all i needed to do was make the footer a little bit smaller in the anon_template so the table isnt cut off

Robin481 (approved):
Amazing fix :roll_eyes: :joy:

PR #827, opened 2025-03-05
Feature 784 certificate master

RandomTannenbaum:


app/helpers/dry_crud/form/builder.rb
```diff
@@ -100,7 +100,7 @@ def float_field(attr, html_options = {})
 
       def decimal_field(attr, html_options = {})
         html_options[:step] ||=
-          (10**-column_property(object, attr, :scale)).to_f
+          (10**-(column_property(object, attr, :scale) || 1)).to_f
```
Robin481, line 103: Ehm why? :grin:

Robin481 (changes requested):
Looks like a good start, please make the Certificates sorted by name though.

Would it be too hard to implement sorting for the table? (e.g. you can press the name header and it will sort by name or you can press the Punkte header and it will sort by Punkte)

app/helpers/dry_crud/form/builder.rb
RandomTannenbaum: I first used the admin view table, thats why I found the bug. When you try to create a decimal_field but the scale property, which is optional, is not defined, it tries to calculate 10**nil, so I used 1 as a default here.

Robin481 (changes requested):
Switch Titel and Bezeichnung please

Robin481 (approved):

PR #843, opened 2025-03-18
Feature/722 people skills controller cleanup

RandomTannenbaum:


app/views/people/people_skills/_form.html.haml
```diff
@@ -1,6 +1,5 @@
 = form_with(model: pskill, url: person_people_skills_path(person)) do |f|
     %div.d-flex.flex-column{data: {controller: "people-skills"}}
-        = f.hidden_field :person_id
```
ManuelMoeri, line 3: Why was this removed?

RandomTannenbaum: The person_id is already in the params and doesn't have to be explicitly added

app/controllers/people/people_skills_controller.rb
```diff
@@ -2,36 +2,54 @@
...
+  layout 'person'
 
   def self.model_class
-    Person
+    PeopleSkill
+  end
+
+  def index
+    rating = params[:rating]
+    return super if rating.present? && ([-1, 0, 1].include? rating.to_i)
```
Robin481, line 19: this seems over complicated doesn't it?

Do I understand it correctly that this is because we want `rating: 0` as a default? Can't you do something with `default_url_params` ?

app/controllers/people/people_skills_controller.rb
```diff
@@ -2,36 +2,54 @@
...
+  end
+
+  def show
+    redirect_to person_people_skills_path(@person)
+  end
+
+  def new
+    super
+    @people_skill.skill ||= Skill.new
+    @category_hidden = @people_skill.skill.category&.id.present? && @people_skill.skill&.id.present?
```
Robin481, line 31: What does this do?

app/javascript/controllers/filter_controller.js
```diff
@@ -9,6 +9,10 @@ export default class extends Controller {
     
         const params = new URLSearchParams(formData);
         const newUrl = `${form.action}?${params.toString()}`;
+
+        const defaultSkillForms = document.querySelectorAll('#default-skills form');
+        defaultSkillForms.forEach(f => f.action = newUrl)
```
Robin481, line 14: This is not understandable when reading the code, maybe a small comment or simplify.

app/views/people/people_skills/_overview.html.haml
```diff
@@ -27,6 +27,5 @@
           =category_child.title + " (#{people_skills_of_category.count})"
           - people_skills_of_category.each do |person_skill|
             - @person_skill = person_skill
-            = form_with model: @person, url: people_skills_person_path(@person, rating: params["rating"]) do |form|
-              = form.fields_for :people_skills, @person_skill do |ff|
-                = render "people/people_skills/edit_form", ff: ff
\ No newline at end of file
+            = form_with model: person_skill, url: person_people_skill_path(@person, person_skill, rating: params["rating"]) do |ff|
```
Robin481, line 30: why use `ff` if you can use a variable that actually says something?

Robin481 (changes requested):
A few small tweaks but nice work!

app/controllers/people/people_skills_controller.rb
RandomTannenbaum: I think it was supposed to hide the category dropdown when selecting an already existing skill, but currently this is already done by a stimulus controller so I think we can remove this

app/controllers/people/people_skills_controller.rb
RandomTannenbaum: I did not find a better solution for this. `default_url_options` specifies the params used when generating urls with url helpers.

Robin481 (approved):

PR #846, opened 2025-03-18
Feature/733 make attr readonly for ptime attrs

Miguel7373:


Gemfile.lock
```diff
@@ -137,8 +137,9 @@ GEM
       railties (>= 6.0.0)
     csv (3.3.2)
     daemons (1.4.1)
-    database_cleaner (2.1.0)
+    database_cleaner (2.0.2)
```
nevio18324, line 140: Twice?

spec/features/people_spec.rb
```diff
@@ -206,6 +206,15 @@ def add_language(language)
       sign_in auth_users(:user), scope: :auth_user
     end
 
+    it 'should have edit fields disabled if ptime sync is active' do
+      allow(Skills).to receive(:use_ptime_sync?).and_return(true)
+      bob = people(:bob)
+      visit person_path(bob)
+      click_link('Bearbeiten', href: edit_person_path(bob))
+      page.find('#person_name').disabled?
+
```
nevio18324, line 215: Would check all inputs

nevio18324 (changes requested):
Maybe also for the translations do you have to use I18n.t ? Maybe something like ti('edit_form.ptime_data') would also work

RandomTannenbaum (changes requested):

Robin481 (changes requested):
Clean up permitted_params as discussed, I'll review it again after that.

Robin481:
Good job :smile_cat:

PR #851, opened 2025-03-24
Rating a unrated skill makes him disappear

nevio18324:


app/controllers/people/people_skills_controller.rb
```diff
@@ -32,8 +32,21 @@ def new
 
   def update
     @people_skills = filtered_people_skills
+    if params['people_skill'][:level] == '0'
+      params['people_skill'][:unrated] = true
```
Robin481, line 36: I don't understand why you set a parameter based on another parameter. And what is it used for?

app/controllers/people/people_skills_controller.rb
```diff
@@ -32,8 +32,21 @@ def new
...
+    end
     super do |format, success|
-      format.turbo_stream { render 'people/people_skills/update', status: :ok } if success
+      update_render(format, success)
+    end
+  end
+
+  def update_render(format, success)
+    if success
+      if params[:rating] == '-1'
```
Robin481, line 45: So the code doesn't really speak for itself, either add a clarifying comment or alternatively stop using magic numbers.

Robin481 (changes requested):
image

app/controllers/people/people_skills_controller.rb
Robin481: Okay sure but in the rest of the application you can't set it back to `0` after having rated it.
E.g. when you list rated skills you can't set them to 0...
Therefore it probably makes more sense to remove the option to set it to `0` after having rated it.

Robin481 (changes requested):

Robin481 (changes requested):
Small adjustments

app/controllers/people/people_skills_controller.rb
```diff
@@ -49,16 +49,24 @@ def show_path
...
 
-    filter_by_rating(@person.people_skills, params[:rating])
+    filter_by_rating(@person.people_skills, params[:rating], current_skill)
   end
 
-  def filter_by_rating(people_skills, rating)
-    return people_skills.where('level > ?', 0) if rating == '0'
-    return people_skills.where(level: 0) if rating == '-1'
+  def filter_by_rating(people_skills, rating, current_skill = nil)
+    if rating == '0'
```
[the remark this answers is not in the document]
nevio18324: that how we always had it its also saved like this in the query

app/controllers/people/people_skills_controller.rb
```diff
@@ -49,16 +49,24 @@ def show_path
...
+      return people_skills.where('level > ?', 0) # Returns all rated skills
+    elsif rating == '-1'
```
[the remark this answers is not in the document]
nevio18324: that how we always had it its also saved like this in the query

app/controllers/people/people_skills_controller.rb
```diff
@@ -49,16 +49,32 @@ def show_path
...
+                            .or(people_skills.where(skill_id: current_skill[:skill_id]))
+      end
+      return people_skills.where(level: 0) # Returns all unrated skills
+    end
+
+    people_skills # If the rating is neither 1 or 0 it returns all
+  end
 
-    people_skills
+  def ratings
```
Robin481, line 73: I didn't say anything about a hash, just use constants...
https://eliotsykes.com/2015/08/18/magic-numbers/

Robin481 (changes requested):

Robin481 (approved):

PR #852, opened 2025-03-25
Feature/787 Add a display_in_cv option

nevio18324:


spec/exporters/odt/cv_spec.rb
RandomTannenbaum, line 1: You are only testing if the educations, member_notes etc. are not displayed when display_in_cv is false but you are not testing the positive case currently.

RandomTannenbaum (changes requested):
**Note**: I would probably put the checkbox on the same line as the label, either in front or after the text. You already did that in the competence notes section. It looks weird to me when there is just a lonely checkbox below the text. Maybe have a chat with UX about this highly sophisticated and philosophical question.

Robin481 (changes requested):
You added a margin between the person notes and the `display_in_cv` label but not in all the other places you use it, please add the margin there as well. 

Otherwise a few tweaks and then we're good :smile:

app/models/activity.rb
```diff
@@ -26,6 +27,7 @@ class Activity < ApplicationRecord
 
   belongs_to :person, touch: true
 
+  validates :display_in_cv, inclusion: [true, false]
```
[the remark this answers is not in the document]
nevio18324: There is a problem with presence that when the value is false presence returns false

db/migrate/20250325145055_add_display_in_cv_option.rb
```diff
@@ -0,0 +1,9 @@
+class AddDisplayInCvOption < ActiveRecord::Migration[8.0]
+  def change
+    tables = [:activities, :advanced_trainings, :educations, :projects]
+    tables.each do |table|
+      add_column table, :display_in_cv, :boolean, default: true
```
Robin481, line 5: Please add the not null validation on the database level here as well and then we are fine :)

db/migrate/20250325145055_add_display_in_cv_option.rb
```diff
@@ -0,0 +1,9 @@
...
+    end
+    add_column :people, :display_competence_notes_in_cv, :boolean, default: true
```
Robin481, line 7: Please add the not null validation on the database level here as well and then we are fine :)

Robin481 (changes requested):

Robin481 (changes requested):
One last small thing:
I think `Im CV anzeigen` is not specific enough, let's call it `Im CV Export inkludieren` because otherwise it might be confusing.
Similarly the `Wird im CV nicht angezeigt` on hover should probably be `Wird im CV Export nicht inkludiert`

Robin481 (changes requested):
Good Job! :balloon:

PR #862, opened 2025-04-07
Feature/848 unify duplicated skills script

nevio18324:


spec/tasks/unify_task_test.rb
```diff
@@ -0,0 +1,16 @@
...
+  describe 'unify' do
+
+    before do
+      Rake.application.rake_require 'tasks/unify'
+    end
+
+    it "should find correct duplicates" do
+      expect {
+        Rake::Task["unify"].execute
+      }.to output("Duplicate of: JUnit id: 677333953 Duplicate: cunit id: 625945042\nDuplicate of: WebComponents id: 1031312693 Duplicate: Web Components id: 157721798\n").to_stdout
```
RandomTannenbaum, line 13: Are we sure the ids always stay the same?

lib/tasks/unify.rake
```diff
@@ -0,0 +1,20 @@
...
+      unless matches.include?(duplicate_sentence(match, skill)) # To check if it was already matched the otherway
+        matches += duplicate_sentence(skill, match)
+      end
+    end
+  end
+  puts matches
+end
+
+def duplicate_sentence(skill, match)
+  "Duplicate of: #{skill.title} id: #{skill.id} Duplicate: #{match.title} id: #{match.id}\n"
```
RandomTannenbaum, line 19:
```suggestion
  "Duplicate of: #{skill.title} with id: #{skill.id}. Duplicate: #{match.title} with id: #{match.id}.\n"
```

Maybe the sentence would be a bit more clear with something like this.

RandomTannenbaum (changes requested):

spec/tasks/unify_task_test.rb
nevio18324: Ye they should because they pass locally and in the workflow

spec/tasks/unify_task_test.rb
```diff
@@ -0,0 +1,16 @@
...
+  describe 'unify' do
+
+    before do
+      Rake.application.rake_require 'tasks/unify'
+    end
+
+    it "should find correct duplicates" do
+      expect {
+        Rake::Task["unify"].execute
+      }.to output("Duplicate of: JUnit with id: 677333953 Duplicate: cunit with id: 625945042.\nDuplicate of: WebComponents with id: 1031312693 Duplicate: Web Components with id: 157721798.\n").to_stdout
```
Robin481, line 13: These IDs might have been static by luck, I would assume as soon as something changes in our fixtures they won't be the same.
Therefore don't use static ids, find the actual models and use their id.

Oh and I also changed your message so the skill is failing anyway :grin:

Robin481 (approved):
Good Job! :smile_cat:

PR #864, opened 2025-04-08
Feature #835 Header and Banner rework using Puzzle Shell 

### LINKED ISSUE #835: Header and Banner rework using Puzzle Shell
Currently the Header and Banner design kinda seems odd compared to other puzzle applications.
To change that we can use the [Puzzle shell](https://github.com/puzzle/puzzle-shell) it offers templates for both header and banner 
For the header we can make it similar to this [example](https://puzzle.github.io/puzzle-shell/?path=/story/puzzle-shell-components-topbar--with-menu-actions)

For the banner we could use something like [this](https://puzzle.github.io/puzzle-shell/?path=/story/puzzle-shell-components-banner--with-one-level-navigation-and-search) without the search

### Todo's ###
- Implement the Puzzle Shell Header with all the features our current header has
- Implement the Puzzle Shell Banner with all the features our current banner has

Miguel7373:


Robin481 (changes requested):
A few small things:
- On smaller screens the language dropdown collides with the logo pretty quickly, I would assume the PuzzleShell has solutions to handle this more gracefully? (two rows or something)
- When making it even smaller all the buttons dissapear and our easter egg is suddenly a white square that appears. Both of these things shouldn't happen.

Robin481:
We don't want this anymore as we don't see the point of the puzzle shell

PR #865, opened 2025-04-11
Store locale in cookies

nevio18324:


app/controllers/application_controller.rb
```diff
@@ -1,14 +1,18 @@
...
   around_action :switch_locale
 
   default_form_builder SkillsFormBuilder
 
   def switch_locale(&)
-    locale = params[:locale] || I18n.default_locale
-    I18n.with_locale(locale, &)
+    param_locale = params[:locale]
+    redirect_to(locale: cookies[:locale] || I18n.default_locale) unless param_locale
+    cookies.permanent[:locale] = param_locale if true?(params[:set_by_user])
```
Robin481, line 14: As `params[:set_by_user]` can only ever by true or nil you don't really need to do param conversion here.
Just check if the param is present or not.

Robin481 (changes requested):
One small adjustment, working perfectly otherwise! :partying_face:

Robin481 (approved):

PR #879, opened 2025-04-29
 Bug 870 translate dropdown data

Miguel7373:


app/helpers/picture_helper.rb
```diff
@@ -1,4 +1,21 @@
 # frozen_string_literal: true
 
 module PictureHelper
+  def language_skill_svg(level)
```
ManuelMoeri, line 4: Why is this located in the PictureHelper?

app/helpers/picture_helper.rb
```diff
@@ -1,4 +1,21 @@
...
+    if matching_skill_level('native', level)
+      return 'muttersprache'
```
ManuelMoeri, line 6: Shouldn't this and 'keine' be uppercase?

app/helpers/person_helper.rb
```diff
@@ -39,7 +39,9 @@ def group_person_skills_by_category(person)
   end
 
   def common_languages_translated
-    I18nData.languages('DE').collect do |language|
+    locale = I18n.locale
+    locale = 'de' if I18n.locale == :'de-CH'
```
ManuelMoeri, line 43: Why is that?

ManuelMoeri (changes requested):
Only did a quick review here.

app/helpers/picture_helper.rb
RandomTannenbaum, line 1: Maybe rename this to language_skill_helper or smth like that?

config/locales/de.yml
```diff
@@ -247,6 +250,9 @@ de:
       add: "%{model} hinzufügen"
       delete: Löschen
       edit: "%{model} bearbeiten"
+  language_skills:
+    native: Muttersprache
+    none: Keine
```
RandomTannenbaum, line 255: Change to "Keines" - Adjust in other languages as needed

RandomTannenbaum (changes requested):
The "Keine" option in the person role level is not being translated yet.
The company dropdown options are not being translated yet.
Do we not want to translate the cities in the export dropdown?

app/helpers/person_helper.rb
```diff
@@ -39,7 +41,9 @@ def group_person_skills_by_category(person)
...
```
Robin481, line 45: I told you to put a comment there

app/helpers/person_helper.rb
```diff
@@ -11,7 +11,9 @@ def nationality_string(nationality, nationality2)
   end
 
   def person_role_string(person_role)
-    "#{person_role.role.name} #{person_role.person_role_level&.level}
+    role_level = person_role.person_role_level&.level
+    role_level = I18n.t('language_skills.none') if role_level == 'Keine'
```
Robin481, line 15: Why use a language skill translation when it comes to the role_level?

app/helpers/person_helper.rb
```diff
@@ -58,7 +62,24 @@ def uneditable_language?(lang)
   end
 
   def language_skill_levels
-    %w[Keine A1 A2 B1 B2 C1 C2 Muttersprache]
+    [
+      [I18n.t('language_skills.none').to_s, 'Keines'],
```
Robin481, line 66: Why does it need to be in here two times?

app/helpers/person_helper.rb
```diff
@@ -58,7 +62,24 @@ def uneditable_language?(lang)
...
+      %w[A1],
+      %w[A2],
+      %w[B1],
+      %w[B2],
+      %w[C1],
+      %w[C2],
+      [I18n.t('language_skills.native').to_s, 'Muttersprache']
```
Robin481, line 73: Why does it need to be in here two times?

Robin481 (changes requested):
I currently get a crash `The asset "language-levels/keine.svg" is not present in the asset pipeline.` stopping me from reviewing this further.

app/helpers/person_helper.rb
Miguel7373: One is for showing, and the other one is the one that is saved from the form.

app/helpers/person_helper.rb
Miguel7373: One is for showing, and the other one is the one that is saved from the form.

PR #885, opened 2025-05-08
736 Delete expertise topics and connections

Miguel7373:


RandomTannenbaum (approved):
LGTM - Just as a sidenote: Do we really want to remove legacy code instead of just removing the activerecord association and the db tables?

Robin481 (approved):

PR #886, opened 2025-05-08
Add categories and sub categories to skills to rate

nevio18324:


app/views/people/people_skills/index.html.haml
```diff
@@ -10,9 +10,19 @@
             = ti("new_skills_to_rate")
             = "(#{not_rated_default_skills(@person).count})"
           = close_action_link person_people_skills_path(@person, no_default_skill_prompt: "")
-        - not_rated_default_skills(@person).each do |person_skill|
-          = form_with model: person_skill, url: person_people_skills_path(@person, rating: params["rating"]) do |form|
-            = render "people/people_skills/edit_form", ff: form
+        - Category.all_parents.each do |category|
+          %div{id: "#{category.title.parameterize}-skills-container"}
+            %div.profile-header.mw-100.border-bottom.mt-4{"data-scroll-target": "scrollItem"}
```
RandomTannenbaum, line 15: Also using the scroll target here confuses the scrolling controller which leads to weird behavior in our jump-to bar on the side.

RandomTannenbaum (changes requested):
I would not show categories that have no default skills:

image

Also maybe it would be good to change the background color of the default skill headers a bit, so when scrolling down they are easier distinguishable from the already rated skills??

The current logic seems to be very inefficient. You fetch all categories and skills per person and then also fetch categories per person etc. This produces way too many db queries. Take advantage of ActiveRecord functions like `includes` to eager load entries beforehand or try to use subqueries.

PR #888, opened 2025-05-09
Feature/884 Team Skill Tracking over time view

ManuelMoeri:
… for easy navigation

### DIFF
--- a/app/javascript/controllers/chart_controller.js
+++ b/app/javascript/controllers/chart_controller.js
@@ -14,8 +14,6 @@ export default class extends Controller {
     Chart.register(annotationPlugin);
     Chart.register(Colors);
 
-    console.log(this.emptychartValue)
-
     const ctx = this.canvasTarget.getContext("2d");
     const chartData = JSON.parse(this.datasetValue);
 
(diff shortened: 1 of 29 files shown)


config/locales/ja.yml
```diff
@@ -164,6 +164,21 @@ ja:
     global:
       search_placeholder: 履歴書を検索...
       search_skills: スキルも検索する
+  date:
```
ManuelMoeri, line 167: Interesting topic
https://smilenihongo.com/months-in-japanese

app/javascript/controllers/chart_controller.js
```diff
@@ -0,0 +1,66 @@
...
+  static values = {
+    dataset: String,
+    emptychart: String
+  }
+
+  connect() {
+    Chart.register(annotationPlugin);
+    Chart.register(Colors);
+
+    console.log(this.emptychartValue)
```
nevio18324, line 17: ?

nevio18324 (changes requested):
Looks about right just don't forget tests.

Robin481:
Really good job on the seeds! :partying_face: 
Didn't fully complete the review yet, but I noticed that the way it is displayed doesn't allow the user to easily digest the teams progress yet. Or at least not always.
Can you also allow the user to switch to a different graph? 
Something like this would be amazing as well
image

config/locales/de.yml
```diff
@@ -218,6 +219,7 @@ de:
       cv_search: CV Suche
       profile: Profil
       skill_search: Skill Suche
+      skills_tracking: Skills tracking
```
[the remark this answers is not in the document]
ManuelMoeri: :de:

PR #890, opened 2025-05-09
Feature/859 team skill tracking over time

RandomTannenbaum:


app/domain/department_skills_snapshot_builder.rb
```diff
@@ -0,0 +1,21 @@
...
+  # The count of skill_levels per skill_id is equal to the count of people that have rated that
+  # skill in a given department.
+
+  def snapshot_all_departments
+    Person.all.group_by(&:department_id).each do |department_id, people|
+      department_skill_levels = PeopleSkill
+                                .where(person_id: people)
+                                .group_by(&:skill_id)
+                                .transform_values do |value|
+        value.pluck(:level)
```
Robin481, line 15: Can't you just use the block syntax here instead of the not so pretty linebreak?

app/domain/department_skills_snapshot_builder.rb
```diff
@@ -0,0 +1,21 @@
+class DepartmentSkillsSnapshotBuilder
+  # This will create snapshots of all departments with their respective skills and levels.
+  # Each snapshot is an instance of DepartmentSkillSnapshot which takes a hash of skills.
+  # This hash has the format:
+  # { <skill_id1> => [skill_level1, skill_level2], <skill_id2> => [skill_level1, skill_level2] }
+  # The count of skill_levels per skill_id is equal to the count of people that have rated that
+  # skill in a given department.
+
+  def snapshot_all_departments
+    Person.all.group_by(&:department_id).each do |department_id, people|
```
Robin481, line 10: What happens to people without a department_id?

db/migrate/20191203161009_create_department.rb
```diff
@@ -39,7 +39,7 @@ def down
     remove_column :people, :department_id, :integer
     rename_column :people, :department_name, :department
 
-    drop_table :departments
+    drop_table :departments, force: :cascade
```
Robin481, line 42: Hmm I don't like when we edit old migrations :grimacing: why is this needed?

db/migrate/20250509105746_create_department_skill_snapshots.rb
```diff
@@ -0,0 +1,10 @@
+class CreateDepartmentSkillSnapshots < ActiveRecord::Migration[8.0]
+  def change
+    create_table :department_skill_snapshots do |t|
```
Robin481, line 3: So I guess your plan is to go off the `created_at` for sorting the snapshots etc.

I was thinking why not add a `month` and `year` variable to the snapshots that more specifically mark them to be from a certain time... or maybe not.

We also can just use the actual `created_at` all the time even when displaying the snapshots... I am just thinking about what happens if the snapshot doesn't work for a while.... I don't know... useless rant over.

Robin481 (changes requested):
Please also add the `delayed_job` model here so it is easier to work with.

app/domain/department_skills_snapshot_builder.rb
RandomTannenbaum: *Theoretically* every Person should have a department since you have to select one in the frontend but since we dont enforce it on activrecord or db level I just filter out people without a department.

PR #916, opened 2025-06-12
Feature 894 Profile Reminder mail

Miguel7373:


app/mailers/reminder_mailer.rb
```diff
@@ -0,0 +1,5 @@
+class ReminderMailer < ApplicationMailer
+  def update_user_reminder_email(person)
+    mail(to: person.email, subject: 'Time to update your profile!')
```
ManuelMoeri, line 3: Maybe replace this with "You haven't updated your profile in one year" so the user knows how regularly he should update it. If you mentioned that in the mail message you can just resolve this and leave it be.

app/views/layouts/mailer.html.haml
```diff
@@ -0,0 +1,21 @@
...
+  %body
+    .container
+      %p
+        Sie haben Ihr Skills-Profil seit einem Jahr nicht mehr aktualisiert.
+      %p
+        Bitte erneuern Sie Ihr Profil.
+        %br/
+        Puzzle Skills
+      %p.footer-note
+        Dies ist eine automatisch generierte Nachricht. Bitte antworten Sie nicht auf diese E-Mail.
```
ManuelMoeri, line 19: Nice work here, just think about the fact that the subject is currently english and the text is german : )
I don't really see a way to translate this since you don't know what person has what primary language.

app/views/people/_form.html.haml
```diff
@@ -81,6 +81,10 @@
             = form.label :shortname
           %tr
             %td= form.text_field :shortname, {class: "mw-100 form-control", **disabled_with_ptime_sync}
+          %th.fw-normal
```
ManuelMoeri, line 84: I don't like how far apart the check_box and the label are but since every field is like this on the profile i guess it's fine.

db/schema.rb
```diff
@@ -166,6 +167,7 @@
     t.integer "department_id"
     t.string "shortname"
     t.boolean "display_competence_notes_in_cv", default: true, null: false
+    t.boolean "reminder_mails_active", default: true
```
ManuelMoeri, line 170: If this stays true definitely mention this in the changelog (for the 3 people that read this) and maybe even write a readme chapter about this so people know where do disable it.

ManuelMoeri (changes requested):
Overall solid job. I wrote down a few things to think about and then you'll be there.

app/controllers/api/people_controller.rb
```diff
@@ -5,7 +5,8 @@ class Api::PeopleController < Api::CrudController
 
   self.permitted_attrs = %i[birthdate location
                             marital_status updated_by name nationality nationality2 title
-                            competence_notes company_id email department_id shortname]
+                            competence_notes company_id email department_id shortname
```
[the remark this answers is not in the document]
Miguel7373: No

app/mailers/reminder_mailer.rb
Miguel7373: Switched to German to match the text of the mail. And the year does get mentioned in the text of the mail.

app/controllers/admin_controller.rb
```diff
@@ -1,7 +1,23 @@
...
     AuthUser
   end
+
+  def toggle_reminder_mails
+    @company.update(company_params) ? head(:ok) : head(:unprocessable_entity)
+  end
+
+  private
+
+  def set_company
```
Robin481, line 15: I don't see why we need to do this.
I think it is cool that you can decide to send reminder mails for each company.
But it doesn't make sense that I as an admin can only edit my company?
Just make it so that the admin view has the settings for all the companies... every admin should be able to edit it for each company.

app/jobs/send_update_profile_reminder_mail_job.rb
```diff
@@ -0,0 +1,19 @@
...
+        ReminderMailer.update_user_reminder_email(person).deliver_now!
+      end
+    end
+  end
+
+  private
+
+  def needs_reminder?(person)
+    person.updated_at < 1.year.ago ||
+    person.people_skills.exists?(['updated_at < ?', 1.year.ago])
```
Robin481, line 17: Hmm this is a funny one.
The person gets updated via `associations_updated_at` for activities and such... which in turn causes and `updated_at` as well :grimacing: 

Maybe we should also use the `associations_updatet_at` for the people skills?

app/jobs/send_update_profile_reminder_mail_job.rb
```diff
@@ -0,0 +1,19 @@
...
+      company = Company.find(person.company_id)
+      if person.reminder_mails_active && company.reminder_mails_active && needs_reminder?(person)
+        ReminderMailer.update_user_reminder_email(person).deliver_now!
+      end
+    end
+  end
+
+  private
+
+  def needs_reminder?(person)
```
Robin481, line 15: So you are gonna send them a mail every week? :grimacing: 
Please don't if you don't want people to get mad at us... there should be some form of cooldown for each person :thinking: 
Or maybe only send it it in the week where the profile hasn't been updated for a year.
`.where(updated_at: 1.year.ago..1.year.ago + 7.days)

Also sending it after a year seems a tad late... Maybe send one after 6 months as well.

app/jobs/send_update_profile_reminder_mail_job.rb
```diff
@@ -0,0 +1,19 @@
+class SendUpdateProfileReminderMailJob < CronJob
+  self.cron_expression = '0 3 * * 7'
+
+  def perform
+    Person.find_each do |person|
```
Robin481, line 5: What purpose does `find_each` serve here?

app/jobs/send_update_profile_reminder_mail_job.rb
```diff
@@ -0,0 +1,19 @@
...
+      company = Company.find(person.company_id)
```
Robin481, line 6: Why not `person.company?`

app/jobs/send_update_profile_reminder_mail_job.rb
```diff
@@ -0,0 +1,19 @@
...
+      if person.reminder_mails_active && company.reminder_mails_active && needs_reminder?(person)
+        ReminderMailer.update_user_reminder_email(person).deliver_now!
```
Robin481, line 8: So I mean we only will be sending a few mails at a time...
But for scaling's sake you might want to use `deliver_later` ? Would allow us to take advantage of our existing delayed_job infrastructure.

app/views/reminder_mailer/update_user_reminder_email.html.haml
```diff
@@ -0,0 +1,10 @@
+%body
```
Robin481, line 1: By looking at the code this mail template doesn't look all to fancy :grimacing: 
Have you considered adding our logo or something?

config/environments/development.rb
```diff
@@ -85,4 +85,16 @@
     # Add warnings directly to the Rails log
     Bullet.rails_logger = true
   end
+
+  # Mailer configuration
+  config.action_mailer.smtp_settings = {
+    address: ENV['EMAIL_ADDRESS'],
```
Robin481, line 91: I don't like the naming too much, `MAIL_SERVER_ADDRESS` `MAIL_SERVER_USERNAME` and `MAIL_SERVER_PASSWORD` seem more fitting

db/migrate/20250616095102_add_reminder_mails_active.rb
```diff
@@ -0,0 +1,8 @@
+class AddReminderMailsActive < ActiveRecord::Migration[8.0]
+  def change
+    add_column :people_skills, :created_at, :datetime
```
Robin481, line 3: https://guides.rubyonrails.org/v6.1/active_record_migrations.html#migration-overview
No need to do that by hand

Robin481 (changes requested):
Still a bit rough.
I also think you can make the admin options more pretty :small_airplane:

app/controllers/admin_controller.rb
Miguel7373: I have now just put it in the companies list. It seems more fitting for the use case.

app/jobs/send_update_profile_reminder_mail_job.rb
```diff
@@ -0,0 +1,29 @@
+class SendUpdateProfileReminderMailJob < CronJob
+  self.cron_expression = '0 3 * * 7'
+
+  def perform
+    find_people_to_remind.find_each do |person|
```
Robin481, line 5: Use `.each`

app/jobs/send_update_profile_reminder_mail_job.rb
```diff
@@ -0,0 +1,29 @@
...
+      ReminderMailer.update_user_reminder_email(person).deliver_later
+    end
+  end
+
+  private
+
+  def find_people_to_remind
+    Person
```
Robin481, line 13: nice query!

app/views/reminder_mailer/update_user_reminder_email.html.haml
```diff
@@ -0,0 +1,18 @@
+%h2 Ihr Skills-Profil wartet auf Sie
+
+%p
+  Hallo,
+%p
+  Ihr Skills-Profil wurde seit über einer langen Zeit nicht mehr aktualisiert. Eine aktuelle Übersicht Ihrer Fähigkeiten hilft uns, passende und spannende Projekte für Sie zu finden.
```
Robin481, line 6: Please use the informal `Du/Dein`

app/views/reminder_mailer/update_user_reminder_email.html.haml
```diff
@@ -0,0 +1,18 @@
...
+%p
+  Nehmen Sie sich einen Moment Zeit, um Ihr Profil auf den neuesten Stand zu bringen.
+
+%a.button{href: "https://skills.puzzle.ch"} Profil jetzt aktualisieren
```
Robin481, line 10: Have you considered that a hardcoded URL isn't that smart? :wink: 
Please figure out a solution to make this dynamically.

db/migrate/20250708065035_add_reminder_mails_active_to_companies.rb
```diff
@@ -0,0 +1,7 @@
+class AddReminderMailsActiveToCompanies < ActiveRecord::Migration[8.0]
+  def change
+    unless column_exists?(:companies, :reminder_mails_active)
```
Robin481, line 3: That should never be needed.
Either we are actually migrating and the column doesn't exist or something went horribly wrong :grimacing:

app/jobs/send_update_profile_reminder_mail_job.rb
Miguel7373: It had a problem, though. In the OR query there was only the year check and not the reminder_mails_active. I had no test that simulated it with one year, so I quickly updated the query for it to check the reminder_mails_active.

Robin481 (approved):

PR #918, opened 2025-06-13
Feature/356 Open Source Contributions

ManuelMoeri:


app/views/contributions/_contribution.html.haml
```diff
@@ -0,0 +1,14 @@
...
+      %div.d-flex.row.pt-3.pb-5
+        %span.col-3.ps-5
+          = date_range_label contribution
+        %span.col.d-flex.flex-column
+          %span.fw-bolder
+            = contribution.title
+          = contribution.reference
+        - unless contribution.display_in_cv
+          %span.col-1.d-flex.justify-content-end.pe-5
+            %div{ 'data-bs-toggle': 'tooltip', 'data-bs-title': t("people.show.is_not_displayed_in_cv"), 'data-controller': 'tooltip'}
```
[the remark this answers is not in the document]
ManuelMoeri: I opened a ticket for this since this is a common bug that also occurs on main #923

config/locales/de.yml
```diff
@@ -113,6 +117,9 @@ de:
       company:
         one: Firma
         other: Firmen
+      contribution:
+        one: Open source Beitrag
```
Robin481, line 121: why is source small

config/locales/de.yml
```diff
@@ -113,6 +117,9 @@ de:
...
+        other: Open source Beiträge
```
Robin481, line 122: why is source small

app/models/contribution.rb
```diff
@@ -0,0 +1,37 @@
...
+#  display_in_cv  :boolean
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+
+class Contribution < ApplicationRecord
+  include DaterangeModel
+
+  after_create :update_associations_updatet_at
+  after_update :update_associations_updatet_at
+  after_destroy :update_associations_updatet_at
```
Robin481, line 24: I realize this is not your making but the typo triggers me so there is this https://github.com/puzzle/skills/issues/924

Robin481 (changes requested):
Further points:
- If I enter a contribution with a link the link should be clickable when viewing it... Be careful about safety though.
- You made the contributions be excluded from the CV without being able to change it, is that intende? :thinking:

ManuelMoeri:
> Further points:
> 
>     * If I enter a contribution with a link the link should be clickable when viewing it... Be careful about safety though.
> 
>     * You made the contributions be excluded from the CV without being able to change it, is that intende? 🤔

That is indeed intende :smile:

Robin481 (approved):
Looks really good now, good job! 
Let's do https://github.com/puzzle/skills/issues/930 as a PR into this PR first so we can merge the whole thing as a package. 
Cheers :v:

ManuelMoeri:
**TODO**
- [ ] Contributions should be included in CV per default
- [ ] Add checkbox for not including contributions on CV export menu
- [ ] CV should not display contributions section when no contributions are present

nevio18324 (approved):
Works nice :100:

Robin481 (approved):

PR #922, opened 2025-06-18
Feature/921 Skill search rework

RandomTannenbaum:


app/domain/filter_params.rb
```diff
@@ -0,0 +1,78 @@
...
+    filters << default_filter if add_row?
+    filters.delete_at(delete_row.to_i) if delete_row
+
+    filters
+  end
+
+  def search_results(filters)
+    return [] unless filters.any?
+
+    skill_ids = filters.transpose.first
```
Robin481, line 31: Have you considered using a hash instead so you could used named keys instead of having to use `.first` ?

app/domain/filter_params.rb
```diff
@@ -0,0 +1,78 @@
...
+                      .select('person_id')
+
+    PeopleSkill.includes(:skill, :person)
+               .where(person_id: filtered_people, skill_id: skill_ids)
+               .group_by(&:person)
+  end
+
+  def filtered_people_skills(filters)
+    filters.map do |skill_id, level, interest|
+      PeopleSkill.where('skill_id = ? AND level >= ? AND interest >= ?', skill_id, level, interest)
```
Robin481, line 45: You could also use (beginless) ranges and then chain multiple .where queries. This way you can also omit passing `skill_id`
e.g.
```
PeopleSkill
     .where(skill_id:)
     .where(level: level..)
     .where(interest: interest..)
```

Not sure if this is that much more readable, but at least spreading it over multiple lines would help...
I'll let you decide :grimacing:

app/domain/filter_params.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  def search_results(filters)
+    return [] unless filters.any?
+
+    skill_ids = filters.transpose.first
+
+    filtered_people = filtered_people_skills(filters)
+                      .group('person_id')
+                      .having('COUNT(id) = ?', skill_ids.length)
```
Robin481, line 35: Is this the best way to do it? :grin:

app/domain/filter_params.rb
```diff
@@ -0,0 +1,78 @@
...
+  def filters_and_results
+    filters = search_filters
+    results = search_results(filters.filter { |f| f.exclude?(nil) })
+    [filters, results]
+  end
+
+  private
+
+  def search_filters
+    unless skill_ids && levels && interests
```
Robin481, line 17: make it a one liner
`return [default_filter] unless skill_ids && levels && interests`

app/domain/filter_params.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  def interests
+    @params[:interest]&.values.presence
+  end
+
+  def constructed_filters
+    skill_ids.map do |id|
+      id.present? ? id.to_i : nil
+    end.zip(levels.map(&:to_i), interests.map(&:to_i))[0..4]
```
Robin481, line 64: Do we really need the `[0..4] `?

app/domain/filter_params.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  def default_filter
```
Robin481, line 67: Why not make it a static variable?

Robin481 (changes requested):
Looking good so far, a few small tweaks.

Additionally:
It doesn't handle it too well when you search for the same skills twice :thinking: 
e.g. It doesn't find someone that has Skill `APT` when searching for it twice.

A simple solution could be to just remove a skill from the dropdwon if it has already been selected. 
Or at least make the feedback sentence mention that you searched for the same skill twice instead of combining them

app/domain/filter_params.rb
RandomTannenbaum: Not necessarily but it prevents the user from adding an infinite number of filters by manually adding query params.

app/domain/filter_params.rb
RandomTannenbaum: For me personally the way with `>=` is more readable 😁.

Robin481 (changes requested):
Lovely!
One last thing which I forgot in the `filter_by_department` ticket is to display the department a person is in next to or below their name.
For  simplicity's sake please do that here so we have one lovely package :smile:

PR #954, opened 2025-07-02
Bug/844 Profile side-nav gets wrong top when a success message is present

ManuelMoeri:


Robin481 (changes requested):
This breaks the sidebar... The missing offset causes the elements to be selected too late.

Robin481 (approved):

PR #958, opened 2025-07-04
Feature/909 Make Skills usable on mobile

ManuelMoeri:
These changes include fixing of these views:
- The header
- The initial landing page
- The CV part of the profile, yet without the editing forms
- The new profile view


Kagemaru (changes requested):
Looks good, but could you add some padding at the bottom, so the buttons are not glued to the end of the page?

Kagemaru (approved):

PR #960, opened 2025-07-07
Feature/943 Skills-tracking: Sort results and add search

ManuelMoeri:
- [x] Implement sorting for department and skills
- [x] Add a searchbar to the skills select
- [x] Styling
- [x] Tests


nevio18324:
Its not a nice solution but works and is the only way to make the width as wide as the widest option work with a slim select

Robin481 (approved):

PR #965, opened 2025-07-10
 feature/946 Setup cancancan

nevio18324:


app/controllers/application_controller.rb
```diff
@@ -35,8 +35,8 @@ def render_unauthorized_not_conf_admin
     render_unauthorized(helpers.conf_admin?)
   end
 
-  def render_unauthorized(unauthorized)
-    return false if unauthorized
+  def render_unauthorized(authorized)
```
Robin481, line 38: pleeeeaaase use better naming and variables this is unnecessarily complicated :joy:

app/views/layouts/application.html.haml
```diff
@@ -35,7 +35,7 @@
                   - else
                     = ti("devise.mockdata")
             -# Admin panel
-            - if conf_admin?
+            - if admin?
```
Robin481, line 38: Why did we change that, I see the comment @ManuelMoeri left but it was never answered :roll_eyes:

Robin481 (changes requested):

Robin481 (approved):
🥍

PR #971, opened 2025-07-15
Feature/966 docker test setup

nevio18324:


Robin481 (changes requested):
`Add a script bin/spec or bin/test that executes the rake spec command in said container` ... please read all the issue tasks  :grimacing: 

Also fix the rspec container always displaying `whoami: cannot find name for user ID 1000` I want a clean test log

Robin481 (changes requested):
Alright looking very good!

One thing:
Please make `bin/test` work even if the container is not running yet. (e.g. `bin/test` also starts the `test` container and the database.

Robin481 (approved):

PR #972, opened 2025-07-15
Feature/933 Display CV-Search multiple keywords

nevio18324:


nevio18324:
i tested if downcase/upcase makes a diffrence and it indeed doesn't

Robin481 (approved):

PR #975, opened 2025-07-16
Add bundle-audit to gemfile and add it to pipeline

nevio18324:


nevio18324:
I wasn't able to test if it works yet

PR #990, opened 2025-07-22
bug/983 language_skills duplicating

nevio18324:


nevio18324:
This is the old version that was removed in #746
it probably isnt the best solution but atleast works for now
and i dont think spending more time to try and fix the new version is worth it.

PR #998, opened 2025-07-24
Add discard to departments so we can keep department_skill_snapshots …

nevio18324:
…even after a department is deleted


ManuelMoeri (approved):
R+ :+1:

PR #1002, opened 2025-07-25
Feature/947 Every person can edit themselves

ManuelMoeri:


db/migrate/20250725075105_add_auth_user_id_to_people.rb
```diff
@@ -0,0 +1,5 @@
+class AddAuthUserIdToPeople < ActiveRecord::Migration[8.0]
+  def change
+    add_column :people, :auth_user_id, :integer, foreign_key: { to_table: :people }
```
RandomTannenbaum, line 3: You should probably also add `belongs_to` and `has_one` to people and auth_users respectively.
See: https://guides.rubyonrails.org/association_basics.html

spec/features/people_spec.rb
```diff
@@ -379,4 +379,29 @@ def add_language(language)
...
+      sign_in auth_users(:user)
+    end
+
+    it 'logged in person can edit their own profile' do
+      ursula = people(:user)
+      visit person_path(ursula)
+
+      click_link('Bearbeiten', href: edit_person_path(ursula))
+      fill_in 'person_title', with: 'Expert at access control'
+      save_button = find_button("Person aktualisieren")
```
RandomTannenbaum, line 394: There is a `click_button` method

RandomTannenbaum (changes requested):

spec/fixtures/people.yml
```diff
@@ -182,4 +183,22 @@ maximillian:
...
+  title: Just a user
+  competence_notes: |
+    Python
+    Go
+    Unity
+  company: firma
+  email: user@example.com
+  department: dev-one
+  shortname: UU
+  auth_user_id: <%= ActiveRecord::FixtureSet.identify(:user) %>
```
nevio18324: auth_user: user should work aswell

Robin481 (changes requested):
Looking quite good.

I like the information about the current permission but I think they could be implemented a bit better:
- Put them below the sidebar? This feels like the right place where it doesn't take up too much space but is still quite visible. Also it doesn't move around the other page elements.
- Display "you have the rights to edit your own profile" or something if the person is on their own profile.
- Maybe add some readme to further explain the permission system? Knowing that you can't edit something is nice but it would be lovely to also know why 😂

Issue #1017, opened 2025-10-31
Bug: Loading skill set can take a lot of time


windrosew:
**Description**

Skill set loads very slowly, especially if many tabs are opened in the browser

Browsers:
FF 144.0 (64-bit), Chrome Version 141.0.7390.122 (Official Build) (64-bit)

How to reproduce: 
- Open https://skills.puzzle.ch/de/skills

Depending on the system load, this can take quite some time

If it is not reproducible yet:
- Sort table

nevio18324:
Maybe just me but for me the same problem also occurs in chrome

Robin481:
> [@Robin481](https://github.com/Robin481) : Loading still takes forever, but I'm not sure if the newest version is already on prod. I see version number 6.0.0 in the application next to the title. In github, version 6.0.0 dates back to August last year.

Yea the problem still occurs, see here: https://chat.puzzle.ch/channel/puzzleSkills?msg=w5Kd4QFEoGZBknAys
The version is still the same because did a prod release without bumping the version in the hopes of it being fixed quickly 😢

Robin481:
Fixed by https://github.com/puzzle/skills/commit/42b298e853bf2ac8bea4c2ebe36e9c6090991459
and partially https://github.com/puzzle/skills/commit/fcb8f253e49355b5308bf5cdca27161dcc8a6fe3

PR #1035, opened 2026-02-03
Fix: switch to one query for all skills instead of one per skill #1017

nevio18324:


Miguel7373 (changes requested):
In my opinion, this solves the loading page issue, but the filters will still be slow since there seems to be an N+1 query in the sort helper. You should probably take a look at this too.

Miguel7373 (approved):

Issue #1024, opened 2026-01-15
Upgrade to Rails 8.1.2


Robin481:
Title says it all

TODO:

- Upgrade the rails version
- Upgrade as many gems as possible without breaking anything 😬 
- Read the documentation of all the intermediate versions to check if we need to consider anything else
- Make sure the tests still work 🎐 
- Don't forget the docker setup

If there are any gems that can't easily be upgraded create follow up tickets.

PR #1036, opened 2026-02-04
Feat: upgrade to rails 8.1.2 #1024

nevio18324:


nevio18324:
# PR Status — Stand: 4.2

## Done
- All the versions are updated
- I have read through all the changes and implemented everything that makes sense for us

## ToDos
- [ ] We are getting some deprecation warnings when running our application they have to be fixed
```
DEPRECATION WARNING: ActiveSupport::Configurable is deprecated without replacement, and will be removed in Rails 8.2.

You can emulate the previous behavior with `class_attribute`.
 (called from <top (required)> at /home/runner/work/skills/skills/config/application.rb:19)
 ```
- [ ] Read through the migration guide as well to ensure we got everything https://guides.rubyonrails.org/8_1_release_notes.html

Issue #1020, opened 2025-11-21
Rest-API for PuzzleSkills


windrosew:
Work in progress

**Description**
As an AI-Application, I want to be able to get information from PuzzleSkills. In a first step, I want to collect all information about members stored in skills. 

Also for AI-Applications, REST-APIs are one approach to retrieve information from other systems. MCP Servers are only suggested, if the endpoint needs to be selected by the AI. However, MCP server can also be generated from the Rest-Interface using OpenAPI.  Therefore, we only need a REST-API for the first approach.

Also, more sophisticated requirements such as an keyword search that relies on indexed data stored in a vector data base might be of interest later on. However, we need first to play with the data until we can specify the requirements.

**Additional context**
We want to automate the tendering process for WTOs. In order to help sales do decide whether to participate in the tendering or not, we want to enrich the tender with puzzle intern information. We what to know which members would fit the requested personal profiles. Thus we need an interface to PuzzleSkills.

**Acceptance Criteria**

- [x] PuzzleSkills provides a REST-API with OpenAPI description
- The REST-API implements the following function:
  - [x] <GET>cvs : pagenated endpoint that returns the json-Object of the cv of all members. The CV includes all information that is also returned on the cv page of a member

**TODO**
- [x] ...
- [x] Tests
- [x] Update CHANGELOG.md if feature is major
- [x] Update Readme.md if necessary

PR #1048, opened 2026-02-11
feat: open rest-api for puzzle skills #1020

Miguel7373:


nevio18324 (commented):
I wasn't able to smoke test the cvs controller but other than that it looks good

Robin481 (changes requested):

PR #1057, opened 2026-02-17
Feature/1021 Allow entering of more CV Data beforehand

schiltpuzzle:


Robin481 (changes requested):
I haven't really looked at the code yet, just tested the functionality.

Please fix the following and then I will have a proper look:

- Set a limit on how long the string can be for two reasons
- - 1) It gets passed via http Query_String, this has a limit of 10240 characters, if you put something longer than this it will break the whole request
- - 2) The formatting makes no sense at all with very long strings. Set a reasonable limit that doesn't break the template.
- Sort of adding to the last point: We talked about switching up the header formatting to allow for somewhat longer strings. Our longest department on Prod is `Funktionsbereiche` which also exists locally, at least make the formatting work with this. 
- Make sure that the anonymized template and the normal template are the same, you handled one of the two by removing the yellow but you didn't adapt the other one.
- We also discussed if you can modify the yellow depending on the input, I am assuming this is not possible but I would've loved a comment elaborating on this.
- We also discussed adding placeholders for the fields, please do that.

schiltpuzzle:
Most of the `Kundenkürzel` and `Projektkürzel` are around 8 characters. I think making the fields bigger would improve readability. I found examples in the ptime, if you want to see for yourself.

Issue #947, opened 2025-06-30
Access control: Every person can edit themselves


ManuelMoeri:
Before doing this ticket, #946 will need to be done. 
This story consists of two parts:

**1.**
First off, find a good way to get and store the LDAP-username. This username is unique and fits the purpose of matching every logged in person to themselves perfectly. The username can either be read from the huge user-hash that Keycloak provides or it could be fetched from the PuzzleTime alongside all the other data. This would mean that the PuzzleTime API would need to be extended (again).

Currently the AuthUser looks like this, seems like we already get a lot of info from keycloak: 
```
#<AuthUser id: 1, uid: "1234", email: "steiner@puzzle.ch", name: "Robin Steiner", last_login: "2025-07-02 11:46:33.434648000 +0000", is_admin: true, created_at: "2024-07-18 07:59:49.571357000 +0000", updated_at: "2025-07-02 11:46:33.435552000 +0000", is_conf_admin: true>
```

**2.**
As the second step, you will need to use the username you just got to match the person correctly. Then make sure that the logged in person has rights to edit their profile.

**TODO**
- [x] Implement as explaind above
- [x] Smoke test
- [x] Tests

ManuelMoeri:
## Progress 29.07.2025 / Why blocked
The functionality of this story has been implemented and simple tests are written as well. The only thing that is keeping it from being completed is the pipeline. Currently there are 70+ tests failing due to the logged in test user not being able to edit various profile accessed in the specs. The solution im proposing here is that we give the tests users that are used in most tests (like bob) the editor role. This role will be implemented in #948. This still gives us the possibility to keep using Ursula User for our access control tests.

This means that this story can not be merged until 948 is merged into this.

PR #1059, opened 2026-02-18
feat: implement editor role #948

### LINKED ISSUE #948: Access control: Create editor and member coach role
After CanCanCan has been setup (#946) the new roles can now be created. These two roles are the 'editor' role and the 'member-coach' role. Feel free to change the names of them if you think it's not fitting enough.

**Editor**
The editor can edit every profile in the application.

**Member-coach**
The member-coach can edit profiles of people that are in the department he is member-coach off.

**TODO**
- [x] Create roles as defined above
- [x] Tests

Miguel7373:


nevio18324 (changes requested):
We are currently using if (editor? etc.) to check for the permissions i did already fix some of them in #947 but make sure that we always use `can?`

nevio18324 (approved):
This has to wait for #947 before it can be merged

nevio18324 (approved):

Robin481 (changes requested):
Please extend the local seeds/setup so that we have a user for each of the roles. 

I want to be able to change the mail in the application_controller.rb to `editor@puzzle.ch` to instantly have the editor role.

Also when using the `user@puzzle.ch` mail I am unable to edit my own profile, not sure why this happens and I think we already had it before.

app/domain/ptime/people_employees.rb
```diff
@@ -111,5 +112,10 @@ def sanitized_role_name(role_name)
     def person_role_level_id_by_role(role)
       PersonRoleLevel.find_by(level: role[:role_level])&.id || PersonRoleLevel.first.id
     end
+
+    def set_auth_user_id_on_person
+      ldap_username = @ptime_employee_attributes[:ldapname]
+      @person.auth_user_id = AuthUser.find_by!(ldap_username:)&.id
```
Robin481, line 118: Hmm does the `AuthUser` already exist at this point or could this throw an error because the person never logged in therefore an `AuthUser` doesn't exist?

db/migrate/20250725075105_add_auth_user_id_to_people.rb
```diff
@@ -0,0 +1,5 @@
+class AddAuthUserIdToPeople < ActiveRecord::Migration[8.0]
+  def change
+    add_reference :people, :auth_user, type: :integer, foreign_key: { to_table: :people }
```
Robin481: Am I confused or is this a wrong foreign key? 🤨 
Shouldn't it be `to_table: :auth_users` ?

app/models/ability.rb
```diff
@@ -1,47 +1,58 @@
-# frozen_string_literal: true
```
Robin481, line 1: Don't remove please :)

spec/features/people_spec.rb
```diff
@@ -373,10 +375,86 @@ def add_language(language)
...
+      expect(page).not_to have_button('Bewerten')
+    end
+  end
+
+  describe 'Editor Access control' do
+    before(:each) do
+      sign_in auth_users(:editor), scope: :auth_user
+    end
+
+    it  'editor can edit other profiles' do
```
Robin481, line 438: This test has the exact same name as the one 10 lines down 😬  rename one of them or if they do the same only keep one

Robin481 (changes requested):
Alright a few first comments.

I am also highly confused, how does anyone get `editor` rights besides setting the flag via the console? 😬

app/domain/ptime/people_employees.rb
Miguel7373: The user gets their `authUser` set the first time they log in. So the only case where there is no user is when the person has never logged into the Skills tool before. In that case, the ID is initially set to `nil` and then gets assigned once the person logs into the tool for the first time. At least that's how I interpret this logic.

Robin481: Please make sure to check this properly and maybe even add a test for this scenarion

Robin481 (changes requested):
A few points:
- If you select `Eddy Editor` from the dropdown and go to `Andreas Admin`'s Profile it says `you are editing your own profile` I am assuming that is something silly with the seeds but i still want that to be more logical. (also if you go to the `Eddy Editor` profile it doesn't say the above when it probably should... actually it just seems like the seeds are set up in a jumbled way because all the profiles from the drodpown seem to have a similar problem)
- When you go from a profile where you get a green `you are allowed to edit` message to a profile where you don't the green message always flashes first before the correct one is displayed.
-

Robin481 (approved):

PR #1068, opened 2026-02-24
Bug/1046 Skillset modal does not close correctly after browser navigation

schiltpuzzle:


nevio18324 (changes requested):
Not really part of this ticket but i think a close animation similar to the open animation would be nice for the modal
rest looks good

Robin481 (changes requested):
Well you fixed one bug but managed to implement another.

Now if you follow the same reproduction step the modal closes as expected and the background isn't grey anymore but you can't scroll the skillset anymore.

Fix it so that the skillset stays scrollable after closing the modal.

PR #1081, opened 2026-03-18
Feat: add sorting to inactive members table #1047

### LINKED ISSUE #1047: Missing sorting in “Inactive People” table
### **Description**
In the **Admin view**, the **Inactive People table** does not support sorting via the table headers, unlike the other people tables in the application.

### **TODO**
- [ ] Add sorting functionality to the Inactive People table headers
- [ ] Write tests to ensure sorting works correctly and remains consistent

nevio18324:


spec/features/unemployed_people_spec.rb
```diff
@@ -1,39 +1,30 @@
...
-      expect(page).not_to have_content(alice.name)
-      Person.where.not(name: [bob.name, alice.name]).each do |person|
-        expect(page).to have_content(person.name)
-      end
-    end
   end
 
-  it 'should delete person after confirming' do
-    visit admin_people_management_path
+  it 'should delete person after confirming', js: true do
```
nevio18324: is the js: true necessary?

nevio18324 (commented):
Looks good overall only some small things and tests are still failing

Robin481 (changes requested):
Hmm I am a bit confused about this pull request, I think we can do better.

There is a lot of duplication between the two controllers, while I understand why this is happening maybe we can consolidate them and just work with a query param instead of having two whole different controllers?
Also I don't understand what the whole `disable_ptime_sync` has to do with this pull request?

schiltpuzzle:
i used `disable_ptime_sync` to test if the table is behaving right. I also think that this method is a good addition for testing everything related to the ptime sync.

app/models/person.rb
```diff
@@ -73,6 +73,12 @@ class Person < ApplicationRecord
 
   scope :unemployed, -> { employed.invert_where }
 
+  scope :not_synced, lambda {
+    return none unless Skills.use_ptime_sync?
```
Robin481, line 77: just do `return unless Skills.use_ptime_sync?`

spec/features/people_management_spec.rb
```diff
@@ -1,57 +1,125 @@
...
+      visit admin_people_management_path(people: 'not_synced')
+
+      within "#person_#{bob.id}" do
+        accept_confirm("Willst du diesen Eintrag wirklich löschen?") do
+          click_button('Löschen')
+        end
+      end
+
+
+      expect(Person.find_by(name: bob)).to be_nil
```
Robin481: This only works by accident as `bob.to_s` returns the person object's name. 
For consistency you should use `bob.name` here as well.

Robin481 (changes requested):

app/models/person.rb
schiltpuzzle: none is useful in this case

Robin481 (approved):

PR #1098, opened 2026-04-01
feat: implement logic to support the new or logic per added skill to search #1013

### LINKED ISSUE #1013: Skill-Search: OR-Search instead of only AND-Search
Use Case:

A Customer wants `Experienced Java Developers` in their project. 
Therefore, we want to search for people that have on of a list of technologies such as: `Java`, `Java EE`, `Java SE`, `Java 9,10`, `Java 21` etc.

TODO:
- Add a Button to the Skills-Search that toggles between AND-Search and OR-Search
- Make the AND-Search as it is now
- Make the OR-Search trigger a different query that selects people with at least one of the technologies
- Think about a way to display the OR-Search in a reasonable way. (for example you might still want to display all the technologies that are being searched for but just display them as empty if the person hasn't rated one of them.)
- Think about a way to maybe sort the results by who has the most of the technologies that are searched for. Or maybe who hast the highest amount of combined stars or combined experience level.
- Write lovely tests

Miguel7373:


Robin481 (approved):
LGTM 😜

PR #1100, opened 2026-04-02
Feat: implement a url as picture option #170

### LINKED ISSUE #170: Fotos als URL
Es wäre nice, wenn ich die Fotos als URL angeben könnte. Dann könnten wir einfach die URL des Fotos von der Puzzle-Website pasten. Achtung: Optional muss auch Upload eines Fotos möglich sein.

Also etwas in dieser Art:

Auswahl_727

nevio18324:


app/models/person.rb
```diff
@@ -111,11 +111,34 @@ def set_default_languages
...
   private
 
   def picture_size
     return if picture.nil? || picture.size < 10.megabytes
 
     errors.add(:picture, :max_size_10MB)
   end
+
+  def url_valid?(picture_url)
+    return true if picture_url.nil? || picture_url.blank?
```
Miguel7373: We don't need the nil? here 
nil.blank? = true

app/models/person.rb
```diff
@@ -111,11 +111,34 @@ def set_default_languages
...
+
+    url = URI.parse(picture_url)
+    return false if url.host.nil?
+
+    host_regex = /^(\w+\.)?#{ENV.fetch('PICTURE_HOST', 'puzzle.ch')}$/
```
Miguel7373: I believe there needs to be a backslash in front of the dot to prevent other characters

Miguel7373 (changes requested):
Some minor changes

Miguel7373 (approved):

app/models/person.rb
```diff
@@ -111,11 +111,34 @@ def set_default_languages
...
     errors.add(:picture, :max_size_10MB)
   end
+
+  def url_valid?(picture_url)
+    return true if picture_url.blank?
+
+    url = URI.parse(picture_url)
+    return false if url.host.nil?
+
+    host_regex = /^(\w+\.)?#{ENV.fetch('PICTURE_HOST', 'puzzle.ch').sub('.', '\.')}$/
```
Robin481: Please make it so we can put multiple URLs in `PICTURE_HOST`

Robin481 (changes requested):
One small request.

Also don't load the picture in the preview either if it is not from a valid url :)

Robin481 (approved):

Issue #1111, opened 2026-04-14
Drop-down to add skills in project technology


lcanobbio:
**Description**
First of all, take a look at this [document](https://codimd.puzzle.ch/lyX_PlJoQkeihCjFhCNmng?both#). What’s in the document is the goal, but this ticket only requires the implementation of the drop-down menu. You can use [this](https://www.npmjs.com/package/slim-select), it’s already being used in our project This means the drop-down menu looks something like the mock-up and has the functionalities shown in the mock-ups. You can create a new attribute on Projects that has a many to many relation to skills. And when you create a new project, you won’t be able to fill in the free-text field at all, instead, you’ll be prompted to enter information directly into the new attribute. When editing a project, the free-text field will still be displayed, but so will the drop-down menu. You cannot save if there is still content in the free-text field, this means you will need to migrate the data.

**Additional context**
***Mock-up***

<img width="557" height="313" alt="Image" src="https://github.com/user-attachments/assets/c73e3f16-8391-4754-8394-284d081eeda0" />

<img width="519" height="341" alt="Image" src="https://github.com/user-attachments/assets/cdb121b0-eed4-45d2-8600-602147a23c13" />

<img width="514" height="70" alt="Image" src="https://github.com/user-attachments/assets/22ae689f-8d8b-4d62-90c7-7d5d0098fbbb" />

[Mock-ups](https://files.puzzle.ch/apps/collectives/BBT-Java-24/UI-8906767)

**TODO**
- [ ] The drop-down menu is designed
- [ ] When adding project, the drop-down menu appears without a free-text field
- [ ] When editing a project, the drop-down menu appears and you need to migrate from the free-text field to the drop-down menu
- [ ] Test

Miguel7373:
# PR Status — Stand: 29.06.2026
### Reminder: Understand the goal of this PR first, then work on it.

## Done
- The Dropdown should now incluede all the features robin wants
- Tests are written

## ToDos
- [x] Fields have to be able to be seached again
- [x] Pipe should green


## Open Questions / Risks (if any)
So the problem is in our search logic. I have started to fix that, but it's nowhere near done. Right now we check associations just on people, but we need to now change that to recursively also check the association of the association if there are some. This can be either done after the refactoring of this ticket or now. 


Any special-case or conditional logic must be documented in the PR with comments on the code segments.

PR #1121, opened 2026-04-17
feat: implement the merging of the department skill snapshots #1109

### LINKED ISSUE #1109: Allow Migrating of Skill-Tracking Data
Sometimes a company may undergo restructuring and move people into new departments.

Sadly this would cause loss of skill-tracking data.

We want to write a tool in the admin interface that allows skill-tracking data to be migrated.

TODO:
- Create an interface for this migration
- Make it possible to select multiple old skill tracking data sets and consolidate them into a new one for a new department
- Tests

Miguel7373:


Miguel7373:
# PR Status — Stand: 21.04.2026
### Reminder: Understand the goal of this PR first, then work on it.

## Done
- Feature is working

## ToDos
- [x] We need to implement a check if the fields are filled.
- [x] Frontend needs some more help for the user to understand what's happening.
- [x] Complete all translations
- [x] Tests 😄 


Any special-case or conditional logic must be documented in the PR with comments on the code segments.

app/views/admin/merge_department_skill_snapshots/_form.html.haml
```diff
@@ -0,0 +1,24 @@
...
+      = form_with model: @form, url: admin_merge_department_skill_snapshots_path do |f|
+        %div.d-flex.justify-content-center.align-content-center.flex-column
+          %div.d-flex.justify-content-between
+            %div{ data: { controller: "dropdown", dropdown_multiple_value: true } }
+              = f.label :old_department_ids, ti("old_departments"), class: "text-secondary"
+
+              = f.select :old_department_ids,
+                options_from_collection_for_select(@departments, :id, :name, params[:old_department_ids]),
+                {},
+                multiple: true, size: 10, class: "skill-tracking-select", data: { dropdown_target: "dropdown" }
```
lcanobbio: I think it would be great if you could select multiple options from the drop-down menu, so you don’t always have to reopen it to add a department. As I described in this ticket https://github.com/puzzle/skills/issues/1111

lcanobbio (commented):
The rest looks good to me, well done.

app/views/admin/merge_department_skill_snapshots/_form.html.haml
Miguel7373: I've completed the part regarding the non-closing dropdown. However, I think the design requested in your ticket wouldn’t be easy to implement and would significantly increase the complexity of the issue. It would likely require more advanced CSS than Bootstrap can comfortably provide, and adding a large custom CSS block is probably not the best approach for this project.

What’s your opinion on this, @Robin481?

app/views/admin/merge_department_skill_snapshots/_form.html.haml
```diff
@@ -0,0 +1,24 @@
+%div.w-100.d-flex.justify-content-center
+  %div.w-50.mt-4
+    %h2.text-center= t('admin.index.merge_department_snapshots')
+    %div.mt-4
+      = form_with model: @form, url: admin_merge_department_skill_snapshots_path do |f|
```
nevio18324: i would rename `@form` to something like `@merge_department_snapshots_form` just so its clear what form it is

config/locales/de.yml
```diff
@@ -153,6 +156,7 @@ de:
       reminder_mail: Aktiviere Erinnerungs E-Mails für die Firma.
       unify_skills: Skills vereinen
       update_people_manually: Personen manuell aktualisieren
+      merge_department_snapshots: Abteilungsskillverlauf zusammenführen
```
nevio18324: Currently the tab is named `Skill Entwicklung` i would rename either one so they have a similar name e.g. `Skill Entwicklungen zusammenführen` or renaming the tab to `Abteilungsskillverlauf` update this for all languages

config/locales/de.yml
```diff
@@ -162,6 +166,9 @@ de:
     unified_skills:
       create:
         success: Skill %{skill1} wurde erfolgreich mit Skill %{skill2} zum neuen Skill %{new_skill} vereint
+    merge_department_skill_snapshots:
+      create:
+        success: Verlauf der Abteilung %{old_department_names} wurde erfolgreich zusammengeführt und an die Abteilung %{new_department_name} übertragen.
```
nevio18324: imo `Verlauf der Abteilung %{old_department_names} wurde auf die Abteilung %{new_department_name} übertragen.` is enough.

Abteilung is always in singular even when merging multiple departments 😢

config/routes.rb
```diff
@@ -44,6 +44,7 @@
       resources :roles
       resources :companies
       resources :unified_skills, only: [:new, :create]
+      resource :merge_department_skill_snapshots, only: [:new, :create]
```
nevio18324: both seem to work but please change it to resources as we are using resources everywhere. You might have to update some routes.

app/controllers/admin/merge_department_skill_snapshots_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def selected_old_departments
+    Department.where(id: @form.old_department_ids)
+  end
+
+  def selected_new_department
+    Department.find(@form.new_department_id)
+  end
+
```
nevio18324: I think you could put those methods into the FormModel

app/controllers/admin/merge_department_skill_snapshots_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  def merge_skills(old_departments, new_department)
```
nevio18324: please update this name

nevio18324 (changes requested):
Some small changes mostly naming.

Idk if it makes sense but maybe don't display already selected departments in the target department select

nevio18324 (changes requested):
LGTM if the color of the flashes get fixed

Miguel7373:
# PR Status — Stand: 19.05.2026
### Reminder: Understand the goal of this PR first, then work on it.

## Done
- The merging works with multiple departments.
- The new department skills get added to the target and don't overwrite.
- There is a message that makes the user confirm his merge.
- We save the snapshots of the target before merging in a history.
- We show this history in a table on the screen.
- Tests for the error messages of the form are written.

## ToDos
- [x] Tests for merging and seeing the error message fail.
- [ ] Tests for the history need to be written.


## Open Questions / Risks (if any)
- The tests have already taken up a lot of time, so maybe @Robin481 could look a bit more into why it fails exactly.


Any special-case or conditional logic must be documented in the PR with comments on the code segments.

Miguel7373:
# PR Status — Stand: XX.XX.XXXX
### Reminder: Understand the goal of this PR first, then work on it.

## Done
- The test has been fixed

## ToDos
- [x] Tests for the history need to be written.
- [x] Write a test that checks the value of the snapshot before and then after a merge.
- [x] Put the multiselect function in the same place as in [this ticket](https://github.com/puzzle/skills/pull/1174).


Any special-case or conditional logic must be documented in the PR with comments on the code segments.

app/controllers/admin/department_merge_history_controller.rb
```diff
@@ -0,0 +1,4 @@
+class Admin::DepartmentMergeHistoryController < CrudController
+  self.nesting = :admin
+  before_action :render_unauthorized_not_conf_admin
```
Robin481: `conf_admin` doesn't exist on main anymore

app/controllers/admin/merge_department_skill_snapshots_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    old_departments, new_department = extract_departments
+    merge_department_snapshots(old_departments, new_department)
+
+    flash.now[:notice] = t('.success',
+                           old_department_names: old_departments.pluck(:name).join(', '),
+                           new_department_name: new_department.name)
+
+    @merge_histories = load_merge_histories
+    respond_to_format
```
Robin481: Outsource another part of the method if you want it to be short and not the respond_to block 😁

app/views/admin/department_merge_history/_index.html.haml
```diff
@@ -0,0 +1,12 @@
+= turbo_frame_tag "department_merge_history" do
+  = ti('department_merge_history.merge_history')
+
+  - if @merge_histories.any?
+    = plain_table(@merge_histories) do |t|
+      - t.attrs(:created_at)
+      - t.attr :target_department_id do |row|
+        = row.target_department.name
```
Robin481: If we properly delete a department this will break, safe operate please

app/controllers/admin/merge_department_skill_snapshots_controller.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+class Admin::MergeDepartmentSkillSnapshotsController < ApplicationController
```
Robin481, line 3: what about some authorization here?

Robin481 (changes requested):
1) Rebase
2) Why not wrap the whole merging into a transaction so that if something fails along the way we have a safe database state 😁

Robin481 (approved):

PR #1123, opened 2026-04-17
Feat: CV Search should show categories the of the keyword #1104

lcanobbio:


app/domain/people_search.rb
Miguel7373: I know this isn't your fault, but I believe it would be nice if this potential N + 1 would get removed.

Miguel7373: Lowkey don't know the purpose of this find_all here. We don't use the output of it, so we could just use each.

app/domain/people_search.rb
```diff
@@ -86,21 +86,31 @@ def in_association(person, sym)
   # rubocop:disable Metrics/AbcSize
   def attribute_in_array(array)
     table_name = table_name_of_attr(array[0])
-    in_attributes = { group: which_group(table_name), attribute: table_name,
-                      keywords_in_attribute: [] }
+
+    all_attributes = []
+
     array.each do |t|
+      in_attributes = { group: which_group(table_name), attribute: table_name,
```
Miguel7373: This has the same name as a method. I would like you to think about the naming of all of this class a bit more. Every second item has a misleading name.

app/domain/people_search.rb
```diff
@@ -86,21 +86,31 @@ def in_association(person, sym)
   # rubocop:disable Metrics/AbcSize
   def attribute_in_array(array)
```
Miguel7373: I believe this thing could be better split to keep up single responsibility. It does a lot.

app/views/cv_search/index.html.haml
```diff
@@ -20,12 +20,12 @@
             %div.w-50.d-flex.mb-2.h-25.mt-2
               = link_to result[:person][:name], person_path(result[:person][:id]), {class: "bg-skills-green w-75 text-decoration-none text-white ps-1 p-2 rounded-1", "data-turbo-frame": "_top" }
             %div.d-flex.flex-column.w-100.h-100.mt-2
-              - result[:found_in].each do |found_in|
+              - result[:found_in].sort_by { |f| [f[:value].first.to_s.length, f[:value].first.to_s.downcase] }.each do |found_in|
```
Miguel7373: You are better off moving this in the controller, I believe. It's a bit too much logic to put in a HAML file in my opinion.

Miguel7373 (changes requested):
Sometimes the filter selects and brings me to the title of a section and not the skill. I saw that this was always the case but found it a pretty annoying thing. I also don't know if there is a ticket already for that. But I would recommend talking with Robin about this and asking if that should be done if a follow-up ticket.

app/domain/people_search.rb
lcanobbio: N + 1 query is removed.

app/domain/people_search.rb
```diff
@@ -11,124 +11,168 @@ class PeopleSearch
...
+      }
     end
-    result
   end
 
-  def in_attributes(attrs)
-    attribute = []
-    searchable_fields(attrs).find_all do |key, value|
-      next if value.nil?
+  def shorten_if_to_long(value, keywords)
```
nevio18324:
```suggestion
  def shorten_if_too_long(value, keywords)
```

nevio18324 (changes requested):

app/domain/people_search.rb
lcanobbio: done

app/domain/people_search.rb
```diff
@@ -11,124 +11,168 @@ class PeopleSearch
...
 
-  def find_matches(people)
-    search_terms.each do |search_term|
-      matches = Person.all.search(search_term)
-      people = matches & people
-      people = matches if search_term == search_terms[0]
+  def find_matching_people
+    matched_people = search_terms.each_with_index.inject([]) do |accumulated_people, (term, index)|
+      current_matches = Person.all.search(term).to_a
+      index.zero? ? current_matches : (accumulated_people & current_matches)
```
Miguel7373:
```suggestion
    matched_people = search_terms.reduce(Person.all) do |scope, term|
      scope.merge(Person.search(term))
    end
```

That is my quick suggestion. I tried to avoid having to query inside a loop. Please test if it does exactly what you want, but in my opinion, it does.

app/domain/people_search.rb
```diff
@@ -11,124 +11,168 @@ class PeopleSearch
...
     end
-    in_attributes[:keywords_in_attribute].length.positive? ? in_attributes : []
+
+    results
   end
 
-  # rubocop:enable Metrics/AbcSize
-  # rubocop:enable Metrics/MethodLength
+  def search_attributes(attributes)
+    searchable_fields(attributes).filter_map do |key, value|
```
Miguel7373, line 128: I believe the filter already takes out the empty values, so you could just delete the check.

app/domain/people_search.rb
```diff
@@ -11,124 +11,168 @@ class PeopleSearch
...
 
-  def in_attributes(attrs)
-    attribute = []
-    searchable_fields(attrs).find_all do |key, value|
-      next if value.nil?
+  def shorten_if_too_long(value, keywords)
+    value.length >= 50 ? shorten(value, keywords) : value
+  end
+
+  def shorten(text, keywords)
```
Miguel7373, line 147: Look into if you can get rid of each double somehow to make this more readable.

Miguel7373 (changes requested):

app/domain/people_search.rb
lcanobbio: I have tested and it works.

app/domain/people_search.rb
lcanobbio: done

Robin481 (changes requested):
I'll give you some functional feedback before reviewing the code:

- The cutoff for when the result is displayed between `... ...` is not yet well defined. If you put `Associate Degree in Health Science and Javascript` as your title and search for `javascript` it will overflow. Make sure that the overflow is handled better.
- The highlighted section puts too much whitespace to the right of the highlighted section making it look like a cut off word 
- I don't like white as the highlighted color as it blends too much with the background. Although I also don't know what would be better. Maybe just make it more grey/off-white.

app/domain/people_search.rb
```diff
@@ -11,124 +11,167 @@ class PeopleSearch
...
 
-  def attribute_not_in_array(target)
-    result = in_attributes(target.attributes)
-    if result.length.positive?
-      result[0][:attribute] = target.class.name.downcase
+      {
+        group: determine_group(key),
+        attribute: key,
+        keywords_in_attribute: matched_keywords,
+        value: [shorten_if_too_long(value, matched_keywords)]
```
Robin481: I don't think this needs to be an array?

app/views/cv_search/index.html.haml
```diff
@@ -19,13 +19,12 @@
...
-            %div.d-flex.flex-column.w-100.h-100.mt-2
-              - result[:found_in].each do |found_in|
+            %div.d-flex.flex-column.w-100.h-100.mt-2{"data-controller"=>"highlight-search-term"}
+              - sort_found_in_items(result[:found_in]).each do |found_in|
                 %div.d-flex.flex-row.mb-2
                   %div.w-25.d-flex.justify-content-start.align-items-center.me-5
                     %div.me-1.w-100.h-100
-                      %p.bg-skills-search-result-blue.text-decoration-none.text-white.ps-1.p-2.rounded-1.text-center.mb-0
-                        = found_in[:keywords_in_attribute].sort_by(&:downcase).join(', ')
+                      %p.found-search-term{style: "white-space: pre;"}= found_in[:value].join(', ')
```
Robin481: Why not move this style into the relevant css class?

Robin481 (changes requested):
I started the code review but then found two other bugs:

- Create a title for an education called `Java Javascript` and then search for `Java Javascrip` -> the highlighting still works perfectly -> then add the `t` and search `Java Javascript` the result won't be highlighted anymore.
- Create a title for an education called `Bla Java Javascript Javaistisch` and search for `Java J` -> the search result breaks

app/domain/people_search.rb
lcanobbio: array is removed

app/views/cv_search/index.html.haml
lcanobbio: I have deleted the CSS class because it's not needed anyway.

app/domain/people_search.rb
```diff
@@ -11,124 +11,167 @@ class PeopleSearch
...
+      # Matches the keyword along with its immediate surrounding context
+      # (up to one word before and one word after).
+      regex = /(?:\S+\s+)?\S*#{escaped_keyword}\S*(?:\s+\S+)?/i
 
-      keywords_in_attribute = keywords_in_attribute(value)
-      if keywords_in_attribute.length.positive?
-        attribute.push({ group: which_group(key), attribute: key,
-                         keywords_in_attribute: keywords_in_attribute })
+      text.scan(regex).map do |match|
+        "... #{match.gsub(/\s+/, ' ').strip} ..."
```
Robin481: Do we have to strip here, shouldn't your regex already be able to handle that?

Robin481: And what does the `.gsub` do here exactly?

app/helpers/cv_search_helper.rb
```diff
@@ -22,4 +22,13 @@ def url_params(found_in)
     { q: query, section_id: found_in[:group],
       rating: found_in_skills?(found_in) ? 1 : nil }
   end
+
+  def sort_found_in_items(items)
+    items.sort_by do |item|
+      [
+        item[:value].first.to_s.length,
```
Robin481: The code does not what it should.
`:value` is most often just a string
Calling `.first` on a string will return the first character therefore your length will always be `1`

app/javascript/controllers/highlight_search_term_controller.js
```diff
@@ -0,0 +1,36 @@
...
+
+        escapedSearchTerm = this.buildSearchPattern(escapedSearchTerm);
+        const regex = new RegExp(`(${escapedSearchTerm})`, "gi");
+
+        elements.forEach((element) => {
+            const userValue = element.textContent.trim();
+
+            const highlightedText = userValue.replace(
+                regex,
+                '<mark class="rounded-1 bg-skills-gray text-dark" style="padding: 1px; border-radius: 1px;">$1</mark>'
```
Robin481: Why do you need border-radius and rounded-1 at the same time, they overwrite each other, pick one.

Also why not create another css class just as `.found-search-term` and use it instead of inlining css?

spec/features/cv_search_spec.rb
```diff
@@ -77,14 +78,44 @@
...
+  within "#skill_#{skill.id}" do
+    find('.icon.icon-pencil').click
+    fill_in 'skill_title', with: name
+    select category, from: 'skill_category_parent'
+    find("input[type='image']").click
   end
 end
 
-def check_search_results(field_name)
+def check_search_results(field_name, person = bob)
```
Robin481: Put that method into the describe block.
You are using bob who is defined in a `let` inside the block, this only works by accident as `let` also leaks onto `self`. If the structure of the test changes bob might not be available anymore. 
There is no advantage to it being outside the describe block so just move it in there.

spec/features/cv_search_spec.rb
```diff
@@ -77,14 +78,44 @@
...
+      expected_url = /#{person_people_skills_path(alice)}.*q=#{first_skill}/
+      expect(page).to have_link(href: expected_url)
+
+      skills_label = Skill.model_name.human.pluralize
+      check_search_results("#{skills_label}/ #{alice.skills.first.parent_category.title}", alice)
+    end
+  end
+end
+
+def rename_skill_with(name, skill, category = 'Software-Engineering')
```
Robin481: This is structurally fine but there is also no advantage to it being outside the describe block, so just move it in there.

app/domain/people_search.rb
```diff
@@ -11,124 +11,167 @@ class PeopleSearch
...
+    end
+
+    people.includes(associations)
   end
 
-  def in_associations(person)
-    association_symbols.map do |sym|
-      in_association(person, sym)
+  def extract_match_data(person)
+    matches_from_attrs = search_attributes(person.attributes)
```
Robin481: Not sure those short forms really help anything.
Why not call those two variables:
`attribute_matches` and `association_matches` ? Feels like AI naming 🤔

Miguel7373:
I still see the source code when I enter "  m" into the search on the new pulled version of this branch. Could you look into that again, @Robin481?

app/domain/people_search.rb
```diff
@@ -11,124 +11,167 @@ class PeopleSearch
...
+    matches_from_assocs = search_associations(person)
+
+    (matches_from_attrs + matches_from_assocs).flatten.compact
+  end
+
+  def search_associations(person)
+    association_symbols.map do |assoc_name|
```
Robin481: Just call it `association_name` no benefit to make it short here.

app/domain/people_search.rb
```diff
@@ -11,124 +11,167 @@ class PeopleSearch
...
     Person.reflections.keys.excluding('company').map(&:to_sym)
   end
 
-  # rubocop:disable Metrics/MethodLength
-  def in_association(person, sym)
-    target = person.association(sym).target
-    return [] if target.nil?
+  def process_association(person, assoc_name)
+    target = person.association(assoc_name).target
+    return [] if target.blank?
```
Robin481: Why return an empty array if you are gonna remove it anyways?

app/domain/people_search.rb
```diff
@@ -11,124 +11,167 @@ class PeopleSearch
...
+      people_skill && !people_skill.unrated?
     end
+  end
 
-    if target.is_a?(Array)
-      attribute_in_array(target)
-    else
-      attribute_not_in_array(target)
+  def process_collection(collection)
+    return [] if collection.empty?
```
Robin481: Again why return an empty array if you are gonna remove it anyways?

app/domain/people_search.rb
```diff
@@ -11,124 +11,167 @@ class PeopleSearch
...
+
+    table_name = table_name_for(collection.first)
```
Robin481: This is the only place you are using `table_name_for` so I don't think you need to extract this logic into a separate method for now.

app/domain/people_search.rb
```diff
@@ -11,124 +11,167 @@ class PeopleSearch
...
+
+    collection.filter_map do |record|
```
Robin481, line 85: Is there a reason you are using `filter_map` ? Can it even return non truthy values ?

Robin481 (changes requested):
- Maybe try to deduplicate the regex logic in the two highlight controllers as they both use the complex regex.
- Adding spaces to the search will populate the url param with `+`, maybe that is normal behaviour but please check that this doesn't cause any issues.
- Add a delay to the search being triggered. It shouldn't already be searching while i am typing. Something like 200ms or so.

app/domain/people_search.rb
lcanobbio: This is correct i removed the unused strip

lcanobbio: The `sub` function replaces only the first occurrence of a pattern with a specified value, whereas `gsub` replaces all occurrences of the pattern with the specified value. Anything that matches the regex is replaced with `' '`.

app/helpers/cv_search_helper.rb
lcanobbio: You are right, the first one is completely wrong. I have removed it.

app/javascript/controllers/highlight_search_term_controller.js
lcanobbio: The thing about the border radius and the rounded part is correct.  I have deleted one of them

spec/features/cv_search_spec.rb
lcanobbio: I have moved the method into the `describe` block

spec/features/cv_search_spec.rb
lcanobbio: Moved it

app/domain/people_search.rb
lcanobbio: renamed

app/domain/people_search.rb
lcanobbio: renamed

app/domain/people_search.rb
lcanobbio: removed

app/domain/people_search.rb
lcanobbio: removed

app/domain/people_search.rb
lcanobbio: I have deleted the method

app/domain/people_search.rb
lcanobbio: This means that when you use `next`, nil is actually returned, but because we don’t want that in our array, the `filter_map` functions filter out all non truthly values. You can't return falsely values.

app/domain/people_search.rb
Robin481: Well I do understand what gsub does but I do not understand what function it serves in terms of your feature :wink:

lcanobbio: The gsub simply replaces multiple spaces and line breaks with a single space. It just occurred to me that this gsub is unnecessary because we already have formatted strings. that's why I deleted it

app/domain/people_search.rb
```diff
@@ -11,124 +11,164 @@ class PeopleSearch
...
+      # Matches the keyword along with its immediate surrounding context
+      # (up to one word before and one word after).
+      regex = /(?:\S+\s+)?\S*#{escaped_keyword}\S*(?:\s+\S+)?/i
 
-      keywords_in_attribute = keywords_in_attribute(value)
-      if keywords_in_attribute.length.positive?
-        attribute.push({ group: which_group(key), attribute: key,
-                         keywords_in_attribute: keywords_in_attribute })
+      text.scan(regex).map do |match|
+        "... #{match} ..."
```
Kagemaru: Conditional '...' if you're at the start or end of the string would help with the visuals a bit.

You can work with MatchData for this:

```ruby
match.begin(0).zero? # Beginning of text
match.end(0) == text.length # End of text
```

Kagemaru (changes requested):

app/domain/people_search.rb
lcanobbio: I have discussed it with @Robin481 and we've agreed that I will create a follow-up ticket to resolve this bug

lcanobbio: @Kagemaru made a suggestion and i have already incorporated it into this ticket

Robin481 (changes requested):
Hmmm it doesn't handle whitespaces well yet.

Searching for `   m` still breaks a lot of the functionality.
While the search results are displayed the highlighting then breaks everything 😬 

Please strip the search term if there are leading whitespaces so that this doesn't happen.

Robin481 (changes requested):
Nope made it worse.
Now the highlighted text gets completely replaced by what you were searching for.
E.g. search for `Mel` and click on Melisandre's Profile and her name will suddenly be `Mel`

Robin481 (approved):
Nice this works really well now, good job! 🥙

PR #1129, opened 2026-04-20
feat: remove all api endpoints and tests #1088

### LINKED ISSUE #1088: Remove old APIs
We kept the old APIs from the Ember.js days here https://github.com/puzzle/skills/tree/master/app/controllers/api

The idea was that we might use them again at some point.

However we recently decided to just rewrite a new API that was more specifically tailored to the relevant use case: https://github.com/puzzle/skills/pull/1048

Therefore it is time to remove all the references to the old APIs.
They were already removed from the routing here: https://github.com/puzzle/skills/commit/5f1c9574242b1515836cc85f9a434d59be2ebd1d
And their tests deactivated here: https://github.com/puzzle/skills/commit/d9dfbfdc9116faa8d8b71453ebfdc0a5a5ea8726

TODO:
- Remove all the old APIs
- Remove all references to the old APIs 
- Remove all tests for the old APIs

nevio18324:


Miguel7373 (changes requested):
Delete the picture controller that was useless and look into what needs to be removed out of the bin folder due to empty scripts and so on.

Miguel7373 (approved):

Robin481 (approved):

PR #1138, opened 2026-04-22
Dark Mode

schiltpuzzle:


app/assets/stylesheets/styles.scss
```diff
@@ -10,6 +10,16 @@ $skills-dark-blue: #1e5a96;
...
 }
 
+.theme-dark {
+  filter: hue-rotate(180deg) invert(0.9);
+}
+
+//Remove theme filter for all images
+.theme-dark img,
+.theme-dark .modal {
+  filter: hue-rotate(180deg) invert(1.1);
```
Miguel7373: You should use percent here and change this to 100% since it's the max value an invert can have.

app/javascript/controllers/theme_controller.js
```diff
@@ -0,0 +1,79 @@
...
+    disconnect() {
+        if (this.select) {
+            this.select.destroy()
+        }
+    }
+
+    changeTheme(theme) {
+        switch(theme) {
+            case "dark":
+                this.enableThemeDark();
```
Miguel7373: Use return to show what this gets returned instead of writing a break.

app/javascript/controllers/theme_controller.js
```diff
@@ -0,0 +1,79 @@
+import {Controller} from "@hotwired/stimulus"
+import SlimSelect from "slim-select"
+
+export default class extends Controller {
+    connect() {
+        const theme = this.getCookieValue();
```
Miguel7373: You should probably write || “light” to not send the case of no cookie into the default case.

app/views/layouts/application.html.haml
```diff
@@ -26,6 +26,9 @@
             -# Language selector
             %li.d-flex.align-items-center.cursor-pointer.border-start-lg.h-100-lg.ps-2.pe-2{"data-controller": "dropdown"}
               = select :i18n, :language, language_selector, {}, class: "form-control", data:{action: "change->dropdown#navigateOnChange"}
+            -# Theme selector
+            %li.d-flex.align-items-center.cursor-pointer.border-start-lg.h-100-lg.ps-2.pe-2
+              = select_tag :theme, options_for_select([[ti('light'), 'light'], [ti("dark"), 'dark'], [ti("system"), 'system']]), data: { controller: "theme" }                                                    |
```
Miguel7373: Please select a system for your strings. You should not be deciding randomly which to use, single or double quotes.

config/locales/de-CH.yml
```diff
@@ -227,6 +227,9 @@ de-CH:
       profile-not-found: Profiu isch nid gfungä wordä.
       valid_percent_range: muess zwüschä 0 u 200 liegä.
   global:
+    light: Häll
```
Miguel7373: Häu

Miguel7373 (commented):
You should think about reverting some other things too, like the green block in the CV search or the stars in the skills search. They have really low contrast.

Also, all your modals, the language drop-down, and some other drop-downs on the skills tracking page are not inverted. 😃

Miguel7373 (changes requested):

nevio18324 (changes requested):
I just noticed a huge lags while using Dark mode can you fix this?

It also flickers around sometimes when scrolling inside the profile.

Robin481:
https://github.com/puzzle/skills/pull/1197 this was merged, therefore this pr here was closed

PR #1147, opened 2026-04-29
fix: update cron expression #1130

### LINKED ISSUE #1130: Maintenance check on reminder mails to update skills profile
We have a reminder mail that gets sent to every skills user if they don't update their skills profile for exactly 6 months or exactly one year. This is great, but we have gotten complaints that the profiles don't get updated enough, so we should test if the mails even get sent with our trap mail. Also, we can think about sending this mail more often. Talk about this with @Robin481 though.

**TODO**
- [ ] Check if the mails get sent
- [ ] Check if we should send the mail more often
- [ ] Make necessary changes to the mail sender job

nevio18324:


nevio18324:
I checked our mail and we apparently have sent :drum: 0!! reminder mails
@Robin481 can you check via the DB if there any members that should receive Mails

i updated the Weekday from 7 to 0 as 7 might not even be supported

Robin481 (approved):

PR #1155, opened 2026-05-04
Feat: Profile as Start Screen #1136

### LINKED ISSUE #1136: Profile as Start Screen
There have been requests to allow the user’s profile to be used as the default start screen. A possible approach is to introduce a setting that lets users choose their preferred start screen. This preference should be persisted using cookies.

**TODO**
- [x] Set profile as start screen
- [x] Tests
- [x] Update Readme.md

lcanobbio:


nevio18324:
This is mostly done other than the tabbar not working correctly

schiltpuzzle (commented):
I noticed when i click on profile in the tab-bar i see the profile selection page. I would prefer to to see my profile and skip this extra step to search my profile. What do you think @Robin481?

app/controllers/application_controller.rb
```diff
@@ -8,9 +8,18 @@ class ApplicationController < ActionController::Base
...
   default_form_builder SkillsFormBuilder
 
+  def start
+    person = helpers.find_person_by_auth_user
+
+    if person
+      redirect_to person_path(person)
+    else
+      redirect_to people_path
+    end
```
Miguel7373:
```suggestion
    redirect_to helpers.find_person_by_auth_user || people_path
```

app/helpers/tab_helper.rb
```diff
@@ -1,39 +1,50 @@
...
 
-  def person_navbar_path
-    extract_path(PERSON_NAVBAR_REGEX)
+  def person_navbar_path(person)
+    extract_path(person_tabs(person))
+  end
+
+  def determine_person_path(person)
+    if person
+      person_path(find_person_by_auth_user)
```
Miguel7373: Why do we call this again here ?

app/helpers/tab_helper.rb
```diff
@@ -1,39 +1,50 @@
...
+    else
+      people_path
+    end
```
Miguel7373:
```suggestion
  person ? person_path(person) : people_path
```

app/helpers/tab_helper.rb
```diff
@@ -1,39 +1,50 @@
...
       { title: ti('navbar.certificates'), path: certificates_path, admin_only: true },
-      { title: ti('navbar.skills_tracking'), path: department_skill_snapshots_path, admin_only: false }
+      { title: ti('navbar.skills_tracking'), path: department_skill_snapshots_path }
     ]
   end
-  # rubocop:enable Layout/LineLength
 
-  def extract_path(regex)
-    request.path.match(regex)&.captures&.join
+  def extract_path(tabs)
```
Miguel7373, line 25: I would rather not chain on the end here for readability. Try to avoid this.

Miguel7373 (changes requested):
Overall pretty good PR

app/helpers/tab_helper.rb
lcanobbio: changed

app/helpers/tab_helper.rb
lcanobbio: Good idea I have taken it on board

app/helpers/tab_helper.rb
lcanobbio: petit malheur
used again `person`

app/controllers/application_controller.rb
lcanobbio: It works. I have changed it

Robin481 (approved):
:+1:

PR #1156, opened 2026-05-04
 tab indexes not working

### LINKED ISSUE #1134: Tab Indexes not working correctly
In some components navigating with Tab is rather odd as an example when rating a Skill "Interest" can not be accessed via Tab.

**TODO**
- [x] Search for more Components where navigating with Tab does not work
- [x] Fix them if there's too much of them create follow-up tickets
- [ ] Update tests

Timebox 6h or follow up tickets

schiltpuzzle:


Miguel7373:
Tab navigation is fixed, but radio button arrow navigation is reversed (left goes right).

schiltpuzzle:
@Miguel7373 i created a new [issue](https://github.com/puzzle/skills/issues/1160) for this behavior

Robin481 (changes requested):
If there are no stars selected and you tab to the stars there will be no indication that you are currently on the stars, try to fix this as well.

Robin481 (approved):

PR #1157, opened 2026-05-04
feat: only use or queries if expert mode is active #1154

### LINKED ISSUE #1154: Skill-Search Expert Mode Toggle Usability Improvements
Clear Query when untoggling expert mode (because if you made an OR query it will stay an OR query after deselecting expert mode but you can't change it)

nevio18324:


Robin481 (approved):
🍖

Issue #1145, opened 2026-04-27
Add Section-Specific Dropdown to CV Search


Miguel7373:
**Description of feature or bug**
We need to introduce a new section selector dropdown within the CV search interface. This will allow users to narrow down their search query to a specific part of the candidate's CV.

To maintain UI consistency, this new dropdown must reuse the existing styling and component logic from the "Department" dropdown currently used in the Skill Search feature. Additionally, the options within this dropdown need to react dynamically to the "Search with skills as well" toggle.

<img width="1549" height="345" alt="Image" src="https://github.com/user-attachments/assets/2a31ef28-f581-41fe-9d8b-aa9416d61653" />

**TODO**
- [x] Implement the new section selector dropdown in the CV Search.
- [x] Implement the state logic to add "Skills" sections based on the "Search with skills as well" button toggle.
- [x] Update Readme.md if necessary
- [x] Tests 🐛

Miguel7373:
I discussed this topic with @lcanobbio, and we came to the conclusion that the best approach would be to use a Slim Select with sections.

<img width="356" height="365" alt="Image" src="https://github.com/user-attachments/assets/6c01369c-1478-4db4-ac87-35ae2ffac7fb" />

This would allow us to clearly differentiate between associations and attributes. That distinction will help us determine where in the logic we need to adjust the attributes being checked.

@lcanobbio and I both have a rough idea of how the logic could be implemented, so it would probably make sense to discuss it with us before starting work on this ticket.

PR #1168, opened 2026-05-18
Feat: Add Section-Specific Dropdown to CV Search #1145

schiltpuzzle:


schiltpuzzle:
# PR Status — 18.05.2026

## Done
- I finished adding the attribute to the person and implementing the drop-down. But at the moment the data for the drop-down is hard coded.

## ToDos
- [x] Find a way to get the Attributes that the `PeopleSearch` class uses to define the `Found_in` field
- [x] Implement the fix

Issue #1140, opened 2026-04-24
CV-Search allow "handle-whitespace" search


Robin481:
As data is sometimes entered inconsistently searching with or without whitespace is necessary.
Example: `RedHat` vs `Red Hat` is one thing that occurs a lot.

We therefore, want to add a "handle-whitespace" search that allows the user to enter the search Term with whitespace `Red Hat` but the app will search for both `Red Hat` and `RedHat`. 

TODO:
- Add checkbox that enables this "handle-whitespace" search
- Only allow checking of checkbox if a whitespace is present
- Add some form of tooltip that explains it
- Visualize search results smartly
- Tests

Miguel7373:
We should do this after merging this [ticket.](https://github.com/orgs/puzzle/projects/5/views/1?pane=issue&itemId=173519108&issue=puzzle%7Cskills%7C1104)

PR #1173, opened 2026-05-20
Feature/1145 add section specific dropdown to cv search

### LINKED ISSUE #1145: Add Section-Specific Dropdown to CV Search

schiltpuzzle:


schiltpuzzle:
# PR Status — 10.06.2026

## Done
- I could make the searchable fields from the `people_search.rb` show up in the drop-down menu has a `optgroup`. At the moment the drop-down does nothing... 

## ToDos
- [x] The drop-down should filter the entries
- [x] tests

schiltpuzzle:
# PR Status — 16.06.2026

## Done
- I finished adding the filter logic that filters by category.

## ToDos
- [x] Refactor the code in `People Search` so its implemented nicely 
- [x] The data in the Drop-down is very inconsistent with what you can actually can search. Find out what you can search and then create a follow up ticket for those things you can not search, because this is out of scope for this ticket.
- [x] The data in the Drop-down is not yet humanized and not translated  
- [x] At the moment there are no test for the feature. Write those ...

lcanobbio:
# PR Status — 23.06.2026

## Done
I’ve improved the People search and it’s now working. I’ve done the translation, but perhaps a test is failing because of a missing translation. I’ve added the cv_search to turbo streams and turbo frames

## ToDos
- [x] Fix tests
- [x] Write new tests
- [x] Rebase
- [x] Peer Review

spec/features/cv_search_spec.rb
```diff
@@ -125,6 +123,37 @@
...
+        check 'search_skills'
+        check_search_results_only_in_one_found_in(data[:field])
+      end
+    }
+
+    it 'should only results with the correct selected categories be found ' do
+      multi_select_from_slim_select('#category-filter', ["Projekte", "Notizen Member"], true)
+      fill_in 'cv_search_field', with: "Rub"
+
+      sleep 0.5
```
schiltpuzzle: This one worked also wihtout `sleep` for me ...

spec/features/cv_search_spec.rb
```diff
@@ -125,6 +123,37 @@
...
+      {field: "Projekte", value: "duc"},
+      {field: "Stationen", value: "Use"},
+      {field: "Weiterbildungen", value: "Cou"},
+      {field: "Contributions", value: "Add"}].each { |data|
+      it 'should only results with the correct selected category be found' do
+        multi_select_from_slim_select('#category-filter', [data[:field]], true)
+        fill_in 'cv_search_field', with: data[:value]
+        
+        # Tests are flaky in firefox
+        sleep 0.5
```
schiltpuzzle: The check waits the default time as it is without a extra `sleep` function. I tested it without it on my machine and it worked. I this doesn't work you could try replacing the `sleep` with `expect(page).to have_field('search_skills')`

config/locales/en.yml
```diff
@@ -209,14 +212,18 @@ en:
         title: Title
         type_of_exam: Type of exam
   cv_search:
+    associations: Associations
```
schiltpuzzle: For the user the naming Associations doesn't make sence. I would prefeer something like `CV Data` or `CV Related Data`

schiltpuzzle (commented):
Feature looks good. Kudos 👍 

Tests and naming need improvement though ...

spec/features/cv_search_spec.rb
lcanobbio: No, it doesn’t work without it because we have a delay. We only send a request about 0.3 seconds after the user has finished typing.

spec/features/cv_search_spec.rb
lcanobbio: Your suggestion doesn't work because we still have a delay search

config/locales/en.yml
lcanobbio: That makes sense. I have changed it.

app/domain/people_search.rb
```diff
@@ -41,19 +68,15 @@ def find_matching_people(handle_whitespaces: false)
   end
 
   def preload_associations(people)
-    associations = [:department, :roles, :projects, :activities,
-                    :educations, :advanced_trainings, :contributions]
-
-    if search_skills
-      associations << :skills
-      associations << :people_skills
```
Robin481, line 49: I think we still want to preload `people_skills` right?

app/helpers/cv_search_helper.rb
```diff
@@ -29,4 +49,11 @@ def sort_found_in_items(items)
       ]
     end
   end
+
+  def search_skills_checkbox(search_skills)
```
Robin481, line 53: I think this bypasses the turbo stream and loads the whole page instead, is that intentional?

app/domain/people_search.rb
```diff
@@ -1,22 +1,49 @@
...
+      requested_associations = ASSOCIATIONS & @categories.map(&:to_sym)
+    end
+
+    use_defaults = requested_personal_details.empty? && requested_associations.empty?
+
+    set_selected_category(requested_personal_details, requested_associations, use_defaults)
+  end
+
+  def set_selected_category(requested_personal_details, requested_associations, use_defaults)
+    @selected_personal_details = use_defaults ? PERSONAL_DETAILS.dup : requested_personal_details
```
Robin481: Why `PERSONAL_DETAILS` here instead of `PERSON_FIELDS` to stay consistent?

Robin481 (changes requested):
I'll look at it more thoroughly again after you look at these issues:

- The search doesn't really work anymore with the departments. A profile has `/dev/one` but searching for `one` doesn't find the department. This currently works on master. 
- Why can we search and find a persons `Funktionen` but we can't select it as a category for the search?

lcanobbio:
- I have been debugging and have found that `pg_search` cannot handle slashes, as in the example. But that’s not a problem, because we don’t have any departments on prod that contain slashes anyway.
-  About the missing category. for me, it is in the drop-down menu under the `opt_group` `CV Data`

app/domain/people_search.rb
lcanobbio: No we no longer need `people skills` because we apply skills to the person, and that way we already get all the skills that person has, in other words, those are `people skills` too.

app/helpers/cv_search_helper.rb
lcanobbio: We have to render the index there because otherwise the dropdown won't update with the skills.

app/domain/people_search.rb
lcanobbio: makes sense

Robin481:
@lcanobbio Well we don't have any departments *at the moment* :wink: 
I think in general we should be able to handle slashes. 
Slashes don't work when there are two in a string.
Example:
`ruby/rails` anywhere will work fine
`/ruby/rails` is a problem.

I think we should try to make that work.

lcanobbio:
I added the `pg_trgm` extension. That way, PostgreSQL can handle the slashes better. What do you think? @Robin481

lcanobbio:
I compared search performance with a large amount of seed data with and without trigram. Unfortunately, for the tested terms, enabling trigram currently degrades load times.

Here are the benchmark results:

| Search Term | Setup | Total Time | ActiveRecord | Queries | GC |
| :--- | :--- | :--- | :--- | :--- | :--- |
| `and` | **With Trigram** | 1174ms | 708.0ms | 112 | 25.4ms |
| `and` | **Without Trigram** | 769ms | 312.7ms | 107 | 68.9ms |
| `ALGOL 58` | **With Trigram** | 1877ms | 639.8ms | 15 | 213.2ms |
| `ALGOL 58` | **Without Trigram** | 1470ms | 370.9ms | 18 | 77.8ms |
| `Cloud Work under pressure...` | **With Trigram** | 1625ms | 682.7ms | 18 | 78.1ms |

**Observations:**
* Database queries (ActiveRecord) take almost twice as long with trigram enabled.

**Suggestion:**
- To improve the performance of the trigram search, we should consider adding a GIN index https://www.postgresql.org/docs/current/gin.html to the relevant columns. What do you think @Robin481 ?

app/controllers/cv_search_controller.rb
```diff
@@ -1,31 +1,73 @@
 # frozen_string_literal: true
 
 class CvSearchController < ApplicationController
+  before_action :prepare_search_data, only: [:index, :search]
   def index
-    @cv_search_results = should_search? ? search_results : []
-    @query_too_short = query.blank? || query.any? { |q| q.strip.length < 3 }
+  end
```
nevio18324: Do we even need this anymore?

app/controllers/cv_search_controller.rb
```diff
@@ -1,31 +1,73 @@
...
+      query,
+      search_skills: @search_skills,
+      categories: params[:category],
+      handle_whitespaces: handle_whitespaces?
+    )
   end
 
   def query
-    params[:q]&.split(',')
+    params[:q].to_s.split(',').map(&:strip).compact_blank
```
nevio18324, line 44: Instead of strip i would remove all whitespaces i highly doubt someone will search for something like
`A B` if they use Handle Whitespaces this would even allow them to search for 2 letter words. Which could result in Lag

lcanobbio:
**Final performance Test**
with a large amount of data only trigram
```ruby
time = Benchmark.measure do
	Person.search(<search_term>).to_a
end
puts "#{time.real.round(3)} Sekunden"
```

Search Term | Time for one search
-- | --
and | 0.533 Sekunden
entwickler | 0.451 Sekunden
meli | 0.478 Sekunden

app/domain/people_search.rb
```diff
@@ -1,22 +1,49 @@
...
+    use_defaults = requested_personal_details.empty? && requested_associations.empty?
+
+    set_selected_category(requested_personal_details, requested_associations, use_defaults)
+  end
+
+  def set_selected_category(requested_personal_details, requested_associations, use_defaults)
+    @selected_personal_details = use_defaults ? PERSON_FIELDS.dup : requested_personal_details
+    @selected_associations = use_defaults ? ASSOCIATIONS.dup : requested_associations
+  end
+
```
nevio18324, line 37:
```suggestion
  def filter_by_category
    if @categories.empty?
      @selected_personal_details = PERSON_FIELDS
      @selected_associations = ASSOCIATIONS
    else
      @selected_personal_details = PERSON_FIELDS & @categories
      @selected_associations = ASSOCIATIONS & @categories.map(&:to_sym)
    end
  end
```

nevio18324: This should have the exact same output  if i am not mistaken

spec/features/cv_search_spec.rb
```diff
@@ -125,6 +123,38 @@
...
+    it 'should only results with the correct selected categories be found ' do
+      multi_select_from_slim_select('#category-filter', ["Projekte", "Notizen Member"], true)
+      fill_in 'cv_search_field', with: "Rub"
+
+      # because of 0.3 search delay
+      sleep 0.3
+
+      check 'search_skills'
+      check_search_results_only_in_one_found_in(["Projekte", "Notizen Member"])
+    end
```
nevio18324, line 153: If we use different selectors it should be possible without sleep

spec/features/cv_search_spec.rb
```diff
@@ -134,13 +164,25 @@ def check_search_results(field_name, person = bob)
     }
   end
 
+  def check_search_results_only_in_one_found_in(field_name)
+    allowed_names = Array(field_name)
+
+    # because of 0.3 search delay
+    sleep 0.3
```
nevio18324: same here

nevio18324 (changes requested):
The new search looks good and seems to work flawless.
I think the new search might even allow searching using a regex try to figure out if it does.
If it works with a regex reopen the handle white space ticket #1140 and update it.

app/controllers/cv_search_controller.rb
lcanobbio: You're right, I deleted it.

app/controllers/cv_search_controller.rb
lcanobbio: Makes sense

app/domain/people_search.rb
lcanobbio: Yeah, totally that does the same thing

spec/features/cv_search_spec.rb
lcanobbio: done

spec/features/cv_search_spec.rb
lcanobbio: done

lcanobbio:
# PR Status — 16.07.2026

## Done
This ticket is actually finished it has already gone through several peer reviews and is now being reviewed by @Robin481 .
## ToDos
- [ ] Conversation resolving
- [ ] Mergen

app/controllers/cv_search_controller.rb
```diff
@@ -1,31 +1,75 @@
...
+  end
+
+  def compact_query
+    params[:q].to_s.gsub(/\s+/, '').split(',').compact_blank
+  end
```
Kagemaru, line 49: Could we use this instead?

It better explains what the compact part is about.

```suggestion
  def compact_query
    query.map { |part| part.gsub(/\s+/, '') }
  end
```

Kagemaru (changes requested):
Bullet warnings:

```
user: kagemaru
GET /de/cv_search/search?q=test&category%5B%5D=
USE eager loading detected
  Person => [:people_skills]
  Add to your query: .includes([:people_skills])
Call stack
  /home/kagemaru/code/pz/skills/reviews/app/domain/people_search.rb:41:in 'Enumerable#uniq'
  /home/kagemaru/code/pz/skills/reviews/app/domain/people_search.rb:41:in 'PeopleSearch#perform_search'
  /home/kagemaru/code/pz/skills/reviews/app/domain/people_search.rb:23:in 'PeopleSearch#initialize'
  /home/kagemaru/code/pz/skills/reviews/app/controllers/cv_search_controller.rb:35:in 'CvSearchController#people_search'
  /home/kagemaru/code/pz/skills/reviews/app/controllers/cv_search_controller.rb:19:in 'CvSearchController#prepare_search_data'
  /home/kagemaru/code/pz/skills/reviews/app/controllers/application_controller.rb:25:in 'ApplicationController#switch_locale'
```

```
user: kagemaru
GET /de/cv_search/search?q=tes&category%5B%5D=
AVOID eager loading detected
  Person => [:department, :roles, :person_roles, :projects, :activities, :educations, :advanced_trainings, :contributions]
  Remove from your query: .includes([:department, :roles, :person_roles, :projects, :activities, :educations, :advanced_trainings, :contributions])
Call stack
  /home/kagemaru/code/pz/skills/reviews/app/domain/people_search.rb:41:in 'Enumerable#uniq'
  /home/kagemaru/code/pz/skills/reviews/app/domain/people_search.rb:41:in 'PeopleSearch#perform_search'
  /home/kagemaru/code/pz/skills/reviews/app/domain/people_search.rb:23:in 'PeopleSearch#initialize'
  /home/kagemaru/code/pz/skills/reviews/app/controllers/cv_search_controller.rb:35:in 'CvSearchController#people_search'
  /home/kagemaru/code/pz/skills/reviews/app/controllers/cv_search_controller.rb:19:in 'CvSearchController#prepare_search_data'
  /home/kagemaru/code/pz/skills/reviews/app/controllers/application_controller.rb:25:in 'ApplicationController#switch_locale'
```


pls fix!

nevio18324:
The Avoid Eager Load was not fixed as it is probably a false positive

app/domain/people_search.rb
```diff
@@ -41,19 +59,20 @@ def find_matching_people(handle_whitespaces: false)
...
 
+    associations_to_load -= [:skills]
     if search_skills
-      associations << :skills
-      associations << :people_skills
+      associations_to_load += [{ skills: [:parent_category, :category] }] << :people_skills
     end
 
-    people.includes(associations)
+    people.includes(associations_to_load)
```
Kagemaru, line 70:
```suggestion
    Bullet.skip do
      people.includes(associations_to_load)
    end
```

Kagemaru (changes requested):

Kagemaru (approved):

PR #1174, opened 2026-05-20
Feature/1111 Drop-down to add skills in project technology

### LINKED ISSUE #1111: Drop-down to add skills in project technology

schiltpuzzle:


schiltpuzzle:
# PR Status — 03.06.2026

## Done
- Finished Feature
- Fixed couple of tests

## ToDos
- [x] Fix remaining 3 Tests 
- [x] Peer Review

app/views/projects/_project.html.haml
```diff
@@ -24,4 +24,7 @@
                             %span.fw-bold.col-12.col-lg-2
                                 = Project.human_attribute_name(:technology)
                             %span.col
-                                = project.technology
+                                - if project.technology.present?
+                                    = project.technology
+                                - else
+                                    = project.skills.map(&:title).join(', ')
```
Miguel7373:
```suggestion
= project.technology.presence || project.skills.map(&:title).join(', ')
```

The method presence is optimal here. We discussed this in a recent tech talk of mine.

Miguel7373 (changes requested):
This PR looks pretty good. There is just no test yet that selects multiple technologies in the Slim select. You should write one of those.

Miguel7373:
Also rebase 😃

schiltpuzzle:
# PR Status — 10.06.2026

## Done
- I finished adding selects that add multiple technologies in the Slim select

## ToDos
- [x] At the moment the pipeline is failing, because of the `bundle-audit` finding and old version of a gem. I didn't see a dependabot pr for this gem...

db/migrate/20260612131400_create_join_table_project_skill.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJoinTableProjectSkill < ActiveRecord::Migration[8.1]
+  def change
+    create_join_table :projects, :skills do |t|
+      # t.index [:project_id, :skill_id]
```
Miguel7373: Why is this commented out? Please remove them.

db/schema.rb
```diff
@@ -219,6 +219,15 @@
     t.index ["role_id"], name: "index_person_roles_on_role_id"
   end
 
+  create_table "project_skills", force: :cascade do |t|
```
Miguel7373: Why are there two of those tables ?

Miguel7373 (changes requested):
Something does not work here. It seems to only save one in the multi-drop-down and nothing more than that. Look into why that happens and fix it.

db/migrate/20260612131400_create_join_table_project_skill.rb
lcanobbio: I have commented them out

db/schema.rb
lcanobbio: the schema needs a update

lcanobbio:
I have fixt the dropdown. it should now display all the skills that have already been selected

Miguel7373 (approved):
Seems good to me.

Robin481 (changes requested):
First a few pieces of functionality feedback before I look at the code in too much detail:

- [x] Why are the skills in the dropdown not sorted alphabetically? Make them sorted
- [x] When navigating via the up and down arrow the already selected skills are not highlighted when navigating over them. This is especially bad if you select the first 5 skills and then want to navigate to the sixth one. I would argue we should just remove skills from the dropdown that are already selected. Being able to remove the via backspace or clicking on the little cross is good enough. 
- [x] After searching for a skill and then pressing enter the search input should be cleared.
- [x] The dropdown has some horribly confusing boundary behaviour. E.g. Open the dropdown via arrow keys, select the third last skill and then the second last skill, press down and enter to deselect and then sometimes the down key goes up and silly stuff like that. 
- [x] When selecting a skill via enter you can then still continue searching for other skills. When you select one with the mouse button the search field loses focus and you have to manually press into it again. 
- [x] Don't force the user to remove the free text field via validation. We should still allow the user to make changes to projects without being forced to migrate to the new skills.

lcanobbio:
# PR Status — 16.06.2026
## Done
- I have resolved all the bugs in @Robin481 comment

## ToDos
- [x] Send this PR into review maybe Peer review at first about the new changes
- [ ] If  the review has been returned, resolve the conversion

Miguel7373 (approved):

Robin481 (changes requested):
- The skills in the dropdown are still not sorted alphabetically for me
- I think last time i reviewed this it was possible to enter the selected skills and delete them via backspace, this now doesn't work anymore. It really should.

Robin481 (approved):
Two things:
- I am happy that backspace deleting now works, however it is in no way visually communicated to the user. Maybe a placeholder next to the already selected skills that says `backspace to delete` ? Maybe the dropdown already supports something like that.
- The workflow `searching for something` -> `typing e.g. Formula` -> `pressing the arrow down` -> `pressing enter to select Formula` feels a bit clunky. Is there a possibility to skip to having press the arrow down and already have the first search result highlighted and only needing to press enter?

Miguel7373:
# PR Status — Stand: 24.06.2026
### Reminder: Understand the goal of this PR first, then work on it.

## Done
- Ive made all the logic work

## ToDos
- [x] Refactor it a bit or add more comments

Any special-case or conditional logic must be documented in the PR with comments on the code segments.

Miguel7373:
# PR Status — Stand: 29.06.2026
### Reminder: Understand the goal of this PR first, then work on it.

## Done
- The Dropdown should now incluede all the features robin wants
- Tests are written

## ToDos
- [x] Fields have to be able to be seached again
- [x] Pipe should green


## Open Questions / Risks (if any)
So the problem is in our search logic. I have started to fix that, but it's nowhere near done. Right now we check associations just on people, but we need to now change that to recursively also check the association of the association if there are some. This can be either done after the refactoring of this ticket or now. 


Any special-case or conditional logic must be documented in the PR with comments on the code segments.

schiltpuzzle (commented):
The code looks good to me, but there are still some things missing or need improvement ...

- [x] You forgot the translations for the drop-down
- [x] I would prefer ` Select a Skill` over `Select Value` as the drop-down Placeholder.
- [x] The location of the text `(Use backspace to deselect)` in the search field is a bit weird, because the feature to deselect skills has nothing to do with the Search function. Maybe you can find a better solution for this. If not, then this is fine.
- [x] When Selecting Skills with arrow keys, every time you select something you will be put back at the start of the list. I don't know if this is the behavior that we want. Check with @Robin481

lcanobbio:
I have a look on this backspace and deselect thing and there isn’t a solution, otherwise you have to make a lot of changes. But I don’t think it’s really that big a problem

app/javascript/controllers/dropdown_controller.js
```diff
@@ -1,76 +1,178 @@
...
+            ).filter(el => !el.classList.contains('ss-hide') && el.style.display !== 'none');
+
+            if (!options.length) return;
+
+            document.querySelectorAll('.ss-option.ss-highlighted')
+                .forEach(el => el.classList.remove('ss-highlighted'));
+
+            const targetIndex = Math.min(index, options.length - 1);
+            options[targetIndex].classList.add('ss-highlighted');
+        }, 100); // We need this to eliminate the highlight form getting of the first object.
```
nevio18324: very confusing comment

app/javascript/controllers/dropdown_controller.js
```diff
@@ -1,76 +1,178 @@
...
+        if (
+            !this.multipleValue || !input || document.activeElement !== input ||
+            event.key !== 'Backspace' || input.value !== '' || !this.orderedSelection.length
+        ) return;
+
+        event.preventDefault();
+
+        // We remove last selected item manually
+        this.slim.setSelected(this.orderedSelection.slice(0, -1));
+    }
```
nevio18324: there should be a simpler way to do this
```suggestion
    handleKeydown(event) {
       const input = this.searchInput;
       if (
          this.multipleValue &&
          input &&
          document.activeElement === input &&
          event.key === 'Backspace' &&
          input.value === '' &&
          this.orderedSelection.length
        ) {
        event.preventDefault();
        // We remove last selected item manually
        this.slim.setSelected(this.orderedSelection.slice(0, -1));
        }
    }
```

nevio18324 (changes requested):
Just some small improvements to the dropdown_controller

app/javascript/controllers/dropdown_controller.js
lcanobbio: i changed it

app/javascript/controllers/dropdown_controller.js
lcanobbio: Nice suggestion I have added it

app/views/projects/_form.html.haml
```diff
@@ -5,8 +5,18 @@
...
       = f.text_area :description, placeholder: "Description", class: "form-control w-100"
-      = f.label :technology
-      = f.text_area :technology, placeholder: "Description", class: "form-control w-100"
+
+      - if entry.technology.present?
+        = f.label :technology
+        %div.bg-warning-subtle.p-2.rounded.mb-2
+          = ti("migration_hint")
+        = f.text_area :technology, placeholder: "Description", class: "form-control w-100"
+      %div{ data: { controller: "dropdown", dropdown_multiple_value: true, dropdown_placeholder_text_value: ti("dropdown_placeholder"), dropdown_hide_selected_value: true, dropdown_clear_text_value: true } }
```
Robin481: Please make this multiline 😂

Robin481: We already set `dropdown_multiple_value: true` here, do we need to set it again on line 19?

app/views/projects/_form.html.haml
```diff
@@ -5,8 +5,18 @@
...
+        = f.label :skill_ids, Skill.model_name.human.pluralize
+        = f.select :skill_ids,
+              options_from_collection_for_select(Skill.order(:title), 'id', 'title', entry.skill_ids),
```
Robin481: Could you use symbols here instead of strings?

app/views/projects/_project.html.haml
```diff
@@ -15,7 +15,7 @@
                 %span.fw-bold.col-12.col-lg-2
                     = Project.human_attribute_name(:technology)
                 %span.col
-                    = project.technology
+                    = ([project.technology] + project.skills.map(&:title)).compact_blank.join(', ')
```
Robin481: Put this in a helper method

app/models/skill.rb
```diff
@@ -21,6 +21,7 @@ class Skill < ApplicationRecord
   has_many :people, through: :people_skills
   belongs_to :category
   has_one :parent_category, through: :category, source: :parent
+  has_and_belongs_to_many :projects
```
Robin481, line 24: Hmm do we need do add `dependent: :destroy` here so the join table gets cleaned up when a project is deleted? Please check that.

Robin481 (changes requested):
The behaviour of the select is really nice now, good job! 

- The select sometimes highlights the uppermost available option after clicking one from the list and it will stay highlighted. This is counter intuitive and shouldn't happen. 
- Line 18 of the projects_spec is now pointless. This will check for `expect(page).to have_content(project.technology)` aka `expect(page).to have_content(nil)` which is always true and leads to a capybara warning. `Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead. /home/rsteiner/code/skills/spec/features/projects_spec.rb:18`
- You made the fixtures have way less content by only adding actually skills to `duckduckgo` and nothing else, all the other fixtures now just have nil. Don't do that. This would also fixe the problem from above if they all had fixtures.

app/models/skill.rb
nevio18324: from what i have read and tested its the default behavior that the join table gets cleaned up

nevio18324:
> * The select sometimes highlights the uppermost available option after clicking one from the list and it will stay highlighted. This is counter intuitive and shouldn't happen.

In my opinion this is okay as it is still focused element and could be selected using enter.
Same happens when you select something via enter the element the mouse is hovering still gets highlighted.

Robin481 (approved):
Alright good job.
Can we continue with the follow-up tickets to this with PRs into this branch so I don't have our master branch held back by a partially completed feature?

PR #1178, opened 2026-06-03
feat: add permissions to rails workflow #1177

### LINKED ISSUE #1177: GitHub CodeQL and CodeScanning Findings
I have enabled some automatic GitHub Code Quality and Security Pipelines.

TODO:
- Look through the findings
- Solve the easy ones within the 4h Timebox
- Make follow-up tickets for the other ones (without leaking the issues 😁 )
- Mark the unnecessary ones as "not an issue" if that is possible

nevio18324:


Miguel7373 (approved):

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class OmniauthCallbacksController < Devise::OmniauthCallbacksController
-  skip_before_action :verify_authenticity_token, only: :keycloak_openid
```
nevio18324, line 4: Me and @Miguel7373  talked about this, I decided to remove it and it doesnt break anything locally but at the same time i think this action was probably skipped for a reason. @Robin481 do you know more about this?

Robin481: Well since you are not using keycloak locally this won't break locally :wink: 
If I remember correctly it is just the request to Keycloak that CSRF is not happy with so that's why this line exists. We can try on staging if you want to but I would assume it will still be needed.

Robin481 (approved):

PR #1181, opened 2026-06-08
feat: cv search should handle whitespaces correctly #1140

### LINKED ISSUE #1140: CV-Search allow "handle-whitespace" search

nevio18324:


nevio18324:
@Robin481 i am not too sure anymore if this ticket makes sense. Obviously it would be a nice feature. But we would probably have to replace pg_search. I wasn't able to find an alternative https://ruby.libhunt.com/pg_search-alternatives that covers everything we need.

schiltpuzzle (commented):
The code looks good 👍, but there are a few missing requirements from the issue that need to be fixed.

As mentioned in the issue, there is an inconsistency with the search behavior. Searching for `RedHat` with the `Ignore Whitespace` option checked should return results for `Red Hat`. This doesn't happen at the moment. Only works for removing the white-space from `Red Hat`, not the other way around. Please check if we want this behavior.

Missing Requirements:

1. Restrict the "Ignore Whitespace" checkbox so it can only be checked if a whitespace is present.

2. Add a tooltip to the checkbox explaining how the feature works.

Robin481:
> As mentioned in the issue, there is an inconsistency with the search behavior. Searching for `RedHat` with the `Ignore Whitespace` option checked should return results for `Red Hat`. This doesn't happen at the moment. Only works for removing the white-space from `Red Hat`, not the other way around. Please check if we want this behavior.

The issue does not say that. It says `we want to add a "handle-whitespace" search that allows the user to enter the search Term with whitespace Red Hat but the app will search for both Red Hat and RedHat.` Therefore only the one way (removing whitespace) is required.

app/domain/people_search.rb
```diff
@@ -145,14 +152,16 @@ def shorten_if_too_long(value, keywords)
   end
 
   def shorten(text, keywords)
+    text = text.gsub("\n", ' ').strip
```
lcanobbio: Why don't you leave a quick comment here explaining what it does, because at first glance it looks very complicated with all those regex

app/domain/people_search.rb
```diff
@@ -172,10 +181,24 @@ def match_to_text(match, text_length)
   end
 
   def extract_matching_keywords(value)
-    normalized_value = value.to_s.strip.downcase
+    value_string = value.to_s
```
lcanobbio: Use the downcase here again

app/domain/people_search.rb
```diff
@@ -172,10 +181,24 @@ def match_to_text(match, text_length)
...
+  end
+
+  def keyword_match?(value, keyword)
+    value.match?(keyword_regex(keyword))
+  end
+
+  def keyword_regex(keyword)
+    if handle_whitespaces
+      escaped = Regexp.escape(keyword.to_s.gsub(/\s+/, ''))
+      Regexp.new(escaped.chars.join('\s*'), Regexp::IGNORECASE)
```
lcanobbio: Remove the `Regexp::IGNORECASE` because we're already doing it at the top

app/domain/people_search.rb
```diff
@@ -172,10 +181,24 @@ def match_to_text(match, text_length)
...
+    else
+      escaped = Regexp.escape(keyword.to_s.strip)
+      Regexp.new(escaped, Regexp::IGNORECASE)
```
lcanobbio: Remove the `Regexp::IGNORECASE` because we're already doing it at the top

lcanobbio (changes requested):
would be nice if that was highlighted too.

Screenshot from 2026-06-16 09-19-14

But the rest is really good

lcanobbio (approved):

Robin481 (changes requested):
I added:
`redhat` in a person's core competences
`redhat` in a person's educations
`red hat` in a station's title
`REDHAT` in a project
When searching for `red hat` and enabling the white space search the first two results are displayed in a broken way.

Robin481 (changes requested):
Looking really good!
One last thing:
Find a better wording than just `Leerzeichen ignorieren` since it actually does more than that (it searched for both it doesn't just ignore them) `Auch ohne Leerzeichen suchen` is the best I can come up with but maybe you have a better Idea.
Or just add a tooltip if that is easier.

nevio18324:
grafik

The tooltip is already here. I just chose `Leerzeichen ignorieren` so it's compact

Robin481 (approved):
Good stuff 🥬

PR #1183, opened 2026-06-08
Draft: Add python script that fixes toc page numbering

Robin481:


Robin481:
This is obviously vibe-coded and not code we really want in our codebase, however the fix actually works. Leaving this open so we can maybe take some inspiration from it down the road.

PR #1191, opened 2026-06-17
feat: change from person to auth_user #1188

### LINKED ISSUE #1188: Refactor: Update `auth_user_id` assignment to use LDAP username instead of email
Currently, `auth_user_id` is assigned by matching email addresses after an OmniAuth login. This is not our long-term strategy. We need to transition to matching based on the `Ldap_username` or `Keycloak_id`.

Additionally, we need to address how users are created and synced. Our PTime sync would set `auth_user_id`, but fails if there is no `AuthUser` with the `Ldap_username` or `Keycloak_id` . Since the `Ldap_username` gets set on the first Keycloak login (handled in `auth_user.rb`), we need to shift the `auth_user_id` assignment logic there.

This ticket also includes a general refactoring of `auth_user.rb to clean up existing technical debt.

Additionally with the next PTime deployment a new Endpoint will be added where we can fetch a user via Ldapname or Keycloak id or Email
 
> [!CAUTION]
> When deploying this a manual PTime sync will probably be mandatory

**TODO**
- [ ] Remove existing logic that sets `auth_user_id` via email matching.
- [ ] Implement logic to set `auth_user_id` via `Ldap_username` upon OmniAuth Keycloak login.
- [ ] Ensure `auth_user_id` is successfully set on a user's first login.
- [ ] Refactor `auth_user.rb`.
- [ ] Add test if possible

nevio18324:


nevio18324:
This obviously does not complete #1188  yet just a little name change so the next person isn't as confused as i was

Robin481 (approved):

PR #1194, opened 2026-06-17
feat(nix): add Nix dev shell and justfile for local development

kreativmonkey:
## What

Adds a reproducible, Nix-based development environment so the project can be built and tested without installing the toolchain globally — as an alternative to the existing Docker Compose setup.

### `flake.nix` — dev shell
Provides a `nix develop` shell with the toolchain pinned to match `.tool-versions` / `docker-compose.yml`:

- **Ruby 4.0** (nixpkgs ships 4.0.5; the project requests 4.0.1 — patch-compatible)
- **Node 24** with **Yarn 4.12** via corepack
- **PostgreSQL 18**
- **just** (task runner)
- Native build/runtime deps for the gems: `libxml2`/`libxslt` (nokogiri), `libffi` & `vips` (ruby-vips), `imagemagick` (mini_magick), `libpq` (pg), `openssl`, `zlib`, `libyaml`

The shell hook keeps all state inside the project (`vendor/bundle` for gems, `tmp/corepack-bin` for the Yarn shim, `tmp/postgres` for a project-local database) and exports the `RAILS_DB_*` / `PG*` variables matching `config/database.yml`.

### `justfile` — task runner
Common tasks, grouped (`just` lists them all):

- **Setup:** `just init` (deps + database), `just deps`
- **Database:** project-local PostgreSQL in `tmp/postgres` — `just db-start` / `db-stop` / `db-status`, plus `db` (prepare), `db-migrate`, `db-reset`, `db-seed`, `psql`
- **Run:** `just dev` (Procfile.dev stack), `server`, `console`, `assets`, `routes`
- **Test & quality:** `just test [path]`, `lint`, `lint-fix`, `brakeman`, `audit`
- **Maintenance:** `just clean`

### `.gitignore`
Ignores `/vendor/bundle` (gems are installed into the project by the dev shell).

## Usage

```bash
nix develop      # enter the shell
just init        # gems + JS deps + database
just dev         # run the app
just test        # run the suite
```

## Notes

- The PostgreSQL cluster uses `trust` auth — it is a local-only dev database.
- On non-NixOS Linux (e.g. Arch) the precompiled native gems work directly; Nix only supplies the toolchain.
- Docker Compose remains fully usable; this is purely additive.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,76 @@
+{
+  description = "Dev shell for the Skills Rails app (Ruby 4.0 / Node 24 / PostgreSQL 18)";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
+
+  outputs = { self, nixpkgs }:
+    let
+      systems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+      forEachSystem = f: nixpkgs.lib.genAttrs systems (system: f nixpkgs.legacyPackages.${system});
+    in
+    {
+      devShells = forEachSystem (pkgs: {
+        default = pkgs.mkShell {
+          # Toolchain pinned to match .tool-versions / docker-compose.yml.
+          # (nixpkgs ships Ruby 4.0.5; the project requests 4.0.1 — patch-compatible.)
+          packages = with pkgs; [
+            ruby_4_0
+            nodejs_24       # bundles corepack, which provides yarn 4.12.0
+            postgresql_18
+            just            # task runner (see ./justfile)
+
+            # Build/runtime deps for native gems
+            pkg-config
+            libxml2         # nokogiri
+            libxslt         # nokogiri
+            libffi          # ffi / ruby-vips
+            libyaml         # psych
+            openssl
+            zlib
+            postgresql_18.lib  # libpq for the pg gem
+            vips            # ruby-vips (loaded at runtime via FFI)
+            imagemagick     # mini_magick (shells out to the CLI)
+          ];
+
+          shellHook = ''
+            # --- Project-local state, nothing installed globally -----------------
+            export PROJECT_ROOT="$PWD"
+            export BUNDLE_PATH="$PROJECT_ROOT/vendor/bundle"
+            export BUNDLE_BIN="$PROJECT_ROOT/vendor/bundle/bin"
+            export GEM_HOME="$BUNDLE_PATH"
+            export PATH="$BUNDLE_BIN:$PATH"
+
+            # Yarn 4.12 via corepack, kept inside the project.
+            # `corepack enable` requires the target dir to exist beforehand.
+            export COREPACK_HOME="$PROJECT_ROOT/tmp/corepack"
+            mkdir -p "$PROJECT_ROOT/tmp/corepack-bin"
+            corepack enable --install-directory "$PROJECT_ROOT/tmp/corepack-bin" >/dev/null 2>&1 || true
+            export PATH="$PROJECT_ROOT/tmp/corepack-bin:$PATH"
+
+            # --- Database connection (matches config/database.yml defaults) ------
+            export RAILS_DB_HOST="127.0.0.1"
+            export RAILS_DB_PORT="5432"
+            export RAILS_DB_USERNAME="skills"
+            export RAILS_DB_PASSWORD="skills"
+
+            # --- Project-local PostgreSQL (managed via the justfile) -------------
+            export PGDATA="$PROJECT_ROOT/tmp/postgres/data"
+            export PGHOST="$PROJECT_ROOT/tmp/postgres/sockets"
+            export PGPORT="5432"
+            export PGUSER="skills"
+
+            cat <<'EOF'
+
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -0,0 +1,76 @@
...
+      forEachSystem = f: nixpkgs.lib.genAttrs systems (system: f nixpkgs.legacyPackages.${system});
+    in
+    {
+      devShells = forEachSystem (pkgs: {
+        default = pkgs.mkShell {
+          # Toolchain pinned to match .tool-versions / docker-compose.yml.
+          # (nixpkgs ships Ruby 4.0.5; the project requests 4.0.1 — patch-compatible.)
+          packages = with pkgs; [
+            ruby_4_0
+            nodejs_24       # bundles corepack, which provides yarn 4.12.0
```
MasterEvarior: Any reason why Yarn is only provided through `nodejs` here? Would adding the `yarn` package work here too, so I do not have to install something imperatively?

flake.nix
```diff
@@ -0,0 +1,76 @@
...
+            pkg-config
+            libxml2         # nokogiri
+            libxslt         # nokogiri
+            libffi          # ffi / ruby-vips
+            libyaml         # psych
+            openssl
+            zlib
+            postgresql_18.lib  # libpq for the pg gem
+            vips            # ruby-vips (loaded at runtime via FFI)
+            imagemagick     # mini_magick (shells out to the CLI)
```
MasterEvarior, line 35: I also needed another dependency:
```suggestion
            pkg-config
            libxml2         # nokogiri
            libxslt         # nokogiri
            libffi          # ffi / ruby-vips
            libyaml         # psych
            openssl
            zlib
            postgresql_18.lib  # libpq for the pg gem
            vips            # ruby-vips (loaded at runtime via FFI)
            imagemagick     # mini_magick (shells out to the CLI)
            graphicsmagick
```

flake.nix
```diff
@@ -0,0 +1,76 @@
...
+          ];
+
+          shellHook = ''
+            # --- Project-local state, nothing installed globally -----------------
+            export PROJECT_ROOT="$PWD"
+            export BUNDLE_PATH="$PROJECT_ROOT/vendor/bundle"
+            export BUNDLE_BIN="$PROJECT_ROOT/vendor/bundle/bin"
+            export GEM_HOME="$BUNDLE_PATH"
+            export PATH="$BUNDLE_BIN:$PATH"
```
MasterEvarior, line 61: I usually set any env vars outside the shell hook because I think it is more readable. Anything you specify in a `devShell` which is not a recognized attribute name (i.e. `shellHook`) is set as an env var: 

```nix
RAILS_DB_HOST="127.0.0.1";
RAILS_DB_PORT="5432";
RAILS_DB_USERNAME="skills";
RAILS_DB_PASSWORD="skills";
```

flake.nix
```diff
@@ -0,0 +1,76 @@
...
+            export RAILS_DB_USERNAME="skills"
+            export RAILS_DB_PASSWORD="skills"
+
+            # --- Project-local PostgreSQL (managed via the justfile) -------------
+            export PGDATA="$PROJECT_ROOT/tmp/postgres/data"
+            export PGHOST="$PROJECT_ROOT/tmp/postgres/sockets"
+            export PGPORT="5432"
+            export PGUSER="skills"
+
+            cat <<'EOF'
```
MasterEvarior: This is more a nitpick than anything else but if you really want this to be reproducable, `cat` should be something like `${lib.getBin pkgs.cat}`. For example, on my system `cat` is simply an alias for `bat` which has other behavior. 

But adding this means potentially downloading `pkgs.cat` simply for this sole reason. I leave it up to you wether you want to go filly reproducable or not 😄

MasterEvarior (changes requested):
First of all: nice to finally see some Nix PRs around here :)  .[This was literally me when I saw this.](https://media2.giphy.com/media/v1.Y2lkPTc5MGI3NjExaXo3a2ZsMXU4N2Y4enMxaHFlamo0am9wbTA3ZHpnZzkxenk4ZDFzOSZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/OfkGZ5H2H3f8Y/giphy.gif)

I think overall it is pretty good. I had at least one missing dependency but otherwise was up and running pretty fast.  I would move all the env vars out of the `shellHook`, as separate variables.

I also think it would be a good idea to add some documentation to the `README.md`. 1-2 lines for each nix and just. Just specifically can also be interesting for ~the unenlightened~ non-nix users.

flake.nix
kreativmonkey: Yes it's working with the `yarn-berry_4` package that ships the version `4.14.1` but that is not the pinned `4.12.0`. The packages are compatible so it is working, but that was the reason why providing it via `nodejs`.

flake.nix
kreativmonkey: I'll replace it with echo ;-)

kreativmonkey:
Have solved your mentions. Please let me know if there is anything else.

MasterEvarior (commented):
The flake looks good to me 👍 
The documentation aswell but I think @Robin481 should also take a short look

Robin481 (approved):

PR #1196, opened 2026-06-18
fix: Resize Fields in Member CV #1132

### LINKED ISSUE #1132: Resize Fields in Member CV
The input fields are currently quite small, which makes it difficult to see and follow the text while typing. To improve usability, all relevant text fields should be resized based on the amount of content they are expected to hold in production.

This applies to fields used when creating or editing entries such as Projects, Education, and similar CV sections.

**TODO**
- [ ] Increase the size of description and other text fields in the CV forms to better match expected content length and improve readability.

IrynaKosiak7:
#1132 was resolved

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -414,6 +414,7 @@ textarea.form-control {
   height: calc(20em + 0.75rem + calc(var(--bs-border-width) * 2));
 }
 
+
 .model-form-container label {
   margin-bottom: 0;
 }
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -410,6 +410,10 @@ a > img {
   padding-top: 0;
 }
 
+textarea.form-control {
+  height: calc(20em + 0.75rem + calc(var(--bs-border-width) * 2));
+}
+
 .model-form-container label {
   margin-bottom: 0;
 }
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -410,6 +410,12 @@ a > img {
   padding-top: 0;
 }
 
+textarea.form-control {
+  min-height: calc(3em + 0.75rem + calc(var(--bs-border-width) * 1.5));
+  field-sizing: content;
+  resize: vertical;
+}
+
 .model-form-container label {
   margin-bottom: 0;
 }
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -411,7 +411,7 @@ a > img {
 }
 
 textarea.form-control {
-  min-height: calc(3em + 0.75rem + calc(var(--bs-border-width) * 1.5));
+  min-height: 3lh;
   field-sizing: content;
   resize: vertical;
 }
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -410,11 +410,6 @@ a > img {
   padding-top: 0;
 }
 
-textarea.form-control {
-  height: calc(20em + 0.75rem + calc(var(--bs-border-width) * 2));
-}
-
-
 .model-form-container label {
   margin-bottom: 0;
 }
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -409,9 +409,6 @@ a > img {
   @extend .mt-1;
   padding-top: 0;
 }
-textarea.form-control {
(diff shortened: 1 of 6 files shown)


nevio18324 (approved):

Robin481:
Heya
Thanks for your PR.
However the Idea was not that the fields are large in every setting.
It makes no sense for the project fields to all be this large.
image
My screen doesn't even fit them all anymore :grin: 

Here's an example from prod that probably is a representable example for their average size:
image

As the ticket states `based on the amount of content they are expected to hold in production.`
So it is not the way to go to increase the size of all fields. This should be done with care from case to case and taking production into consideration.

Robin481 (changes requested):
see comment above

Robin481 (changes requested):
Heya, this is looking a bit better, however the fields should also already be the appropriate size when loading the page.
Currently when entering a lot of text and then saving and reopening the field will be small again and only when typing something it is resized again.

app/views/activities/_form.html.haml
```diff
@@ -2,9 +2,25 @@
...
-      = f.text_area :role, placeholder: "Rolle", class: "form-control w-100"
+      = f.text_area :role, placeholder: "Rolle", class: "form-control w-100", oninput: "this.style.height = 'auto'; this.style.height = this.scrollHeight + 'px';"
+      :javascript
+        (function() {
+          var el = document.getElementById('activity_role');
+          if (el) {
+            el.style.height = 'auto';
+            el.style.height = el.scrollHeight + 'px';
+          }
+        })();
```
Miguel7373:
```suggestion
      = f.text_area :title, placeholder: "Description", class: "form-control w-100", data: { controller: "autogrow", action: "input->autogrow#resize" }
```
Here’s how you can use the controller we have and trigger an action to resize it live.

Miguel7373 (changes requested):
This is a good start 👍

I’d suggest removing the duplicated inline JavaScript and moving the textarea auto-resize logic into a Stimulus controller. This will make the behavior reusable and improve readability.

Instead of inline JS and the separate script block, you could create a Stimulus controller in app/javascript/controllers, for example:
```js
import { Controller } from "@hotwired/stimulus"

export default class extends Controller {
    connect() {
        this.resize()
    }

    resize() {
        this.element.style.height = "auto"
        this.element.style.height = `${this.element.scrollHeight}px`
    }
}
```
Then you can attach it directly in the view, via a data-controller and input action on the textarea.

This keeps the markup cleaner and avoids inline JavaScript, while still preserving the same behavior.

Miguel7373:
Even though we were all hoping for a non-JS solution, our options are pretty limited since [field-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/field-sizing) isn’t supported in Firefox.

Miguel7373:
As I just found out (though it’s not documented on the website), it works in Firefox v152. That means we could actually use it in the code. @Robin481 what do you think should we relying on something that only works in the latest version of Firefox, or do you prefer sticking with the older approach?

Robin481:
> As I just found out (though it’s not documented on the website), it works in Firefox v152. That means we could actually use it in the code. @Robin481 what do you think should we relying on something that only works in the latest version of Firefox, or do you prefer sticking with the older approach?

Haha the question is what happens in older Firefox Versions? If the "worst case" is that it works as it did before this feature I would argue this is completely fine.

IrynaKosiak7:
@Miguel7373  @Robin481 That’s a new feature; I’ve had a look at it too, but it didn’t work for me (Firefox v151), so I looked for another solution. 
And now I’m waiting to make a final decision – should I use 'field-sizing' in CSS and sort out the other fixes, or leave it in the JS code but refactor it?

Miguel7373:
I think we should use `field-sizing` here. This isn't a large project, so I think it's reasonable to expect Firefox users to be on a recent version. They would only need Firefox 152 if they are even using Firefox.

It also doesn't break any existing behavior. The experience remains the same as it was before the ticket. Given that, this seems like a win.

app/assets/stylesheets/styles.scss
```diff
@@ -382,6 +410,12 @@ a > img {
   padding-top: 0;
 }
 
+textarea.form-control {
+  min-height: calc(3em + 0.75rem + calc(var(--bs-border-width) * 1.5));
```
Miguel7373: I would use just 2 or 3 lh here, depending on what looks better. This would also make it more readable.

Miguel7373 (changes requested):
This looks pretty clean overall now. 👍🏻

Miguel7373 (approved):

Robin481 (approved):
Lovely 🥳

PR #1197, opened 2026-06-18
feat: add dark mode with system-preference default close #1029

### LINKED ISSUE #1029: Dark Mode
Title says it all.

TODO:
- Start implementing a dark mode (figure out the most efficient way of doing this)
- Add a dark mode toggle to the top right of the website
- Make sure all relevant pages handle this gracefully.
- Save the preferred option in a cookie 

Timeboxed:
- As this is not a high priority feature I would like to limit work on this to ~6 Hours and then reevaluate where we stand.

kreativmonkey:
Add a three-state theme switcher (auto / light / dark) in the top-right navbar. The choice is persisted in the `theme` cookie; on first visit the operating system's prefers-color-scheme is used. An inline head script applies the resolved theme before first paint to avoid a flash.

Built on Bootstrap 5.3's native color modes (`data-bs-theme` on <html>), so most components adapt automatically. A soft blue-gray dark palette (not pure black) is layered on top via CSS variable overrides; the Puzzle brand blue header/topbar is intentionally kept in both modes.

- Add ThemeHelper to resolve the cookie into the server-rendered data-bs-theme attribute
- Add a Stimulus `theme` controller that cycles the preference, writes the cookie, follows live OS changes in auto mode, and renders inline SVG icons (no icon-font dependency)
- Swap fixed-color utility classes (bg-white/bg-light/text-dark/text-black) for theme-aware equivalents across views
- Brighten fixed-color image icons in dark mode (CRUD .icon-* backgrounds and standalone .theme-icon  tags) via a lightness-inverting filter
- Add the `global.theme.toggle` label to all six locales


nevio18324 (approved):
Very nice 👍

app/javascript/controllers/theme_controller.js
```diff
@@ -0,0 +1,52 @@
...
+// Cycles the colour scheme preference auto -> light -> dark, persists it in the
+// `theme` cookie and applies it via the <html data-bs-theme> attribute.
+const PREFS = ["auto", "light", "dark"];
+
+// Inline SVGs (Bootstrap Icons, MIT) so the toggle has no icon-font dependency:
+// circle-half = auto, sun-fill = light, moon-stars-fill = dark.
+const svg = (paths) =>
+    `<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="currentColor" viewBox="0 0 16 16">${paths}</svg>`;
+const ICONS = {
+    auto: svg('<path d="M8 15A7 7 0 1 0 8 1zm0 1A8 8 0 1 1 8 0a8 8 0 0 1 0 16"/>'),
```
Robin481: We don't usually use path based images we just have them in our source: https://github.com/puzzle/skills/tree/master/app/assets/images

If you could find appropriately licensed sources to add to our repo that would be lovely.

app/views/layouts/application.html.haml
```diff
@@ -1,6 +1,16 @@
 !!!
-%html
+%html{"data-bs-theme": theme_preference}
   %head
+    -# Resolve the theme before first paint to avoid a flash of the wrong color scheme.
```
Robin481, line 4: I wasn't able to reproduce the issue this code fixes.
I never see a flash of the wrong color scheme.
Can you either show me how to reproduce or remove this if it is actually not needed?

Robin481 (changes requested):
Nice!

In addition to the code changes can you please adjust the following:
- Center the theme-icon in between the two lines. 
- Could you outsource the dark mode changes into their own file so they are more isolated and can be more easily identified. (something like `dark_mode.scss`)
- Could you please also add tests for this feature. 
- When looking at department_snapshots http://localhost:3000/de/department_skill_snapshots in light mode and then switching to dark it won't automatically make the charts use the correct styling and therefore make the information text unreadable. A reload fixes this. Same thing also happens going from dark to light.

app/views/layouts/application.html.haml
kreativmonkey: how do you want to reproduce something that isn't there while it is fixed before the issue comes visible? Like i can see in the Issue #1029 , the current work has this issue too, of flashing the wrong color scheme. 

Without that code, the preferred color will be first initialized while loding the `Stimulus` in theü `theme_controller.js`, which is done near the end of the page load. With these lines, the right color Theme will be selected before the browser paints anything.  

Reproducing:

- Remove these lines
- Set your OS to dark mode
- Clear the cookie for the testpage
- Throttle the network so  the JS bundle loads slowly
- Hard-reload the page

You will see the light theme first, then snap to dark once theme_controller.js is connected.

app/javascript/controllers/chart_controller.js
kreativmonkey, line 1: Solving the issue on the department_skill_snapshots page while switching the color Theme. Now the table will directly repainted on the color switch action.

kreativmonkey:
The following adjustments are done:

- Center the theme-icon in between the two lines.
- Could you outsource the dark mode changes into their own file so they are more isolated and can be more easily identified. (something like dark_mode.scss)
- Could you please also add tests for this feature.
- When looking at department_snapshots http://localhost:3000/de/department_skill_snapshots in light mode and then switching to dark it won't automatically make the charts use the correct styling and therefore make the information text unreadable. A reload fixes this. Same thing also happens going from dark to light.

One question about the centered icon: All the other elements are not centered, why centering these?

Robin481:
> The following adjustments are done:
> 
> * Center the theme-icon in between the two lines.
> * Could you outsource the dark mode changes into their own file so they are more isolated and can be more easily identified. (something like dark_mode.scss)
> * Could you please also add tests for this feature.
> * When looking at department_snapshots http://localhost:3000/de/department_skill_snapshots in light mode and then switching to dark it won't automatically make the charts use the correct styling and therefore make the information text unreadable. A reload fixes this. Same thing also happens going from dark to light.
> 
> One question about the centered icon: All the other elements are not centered, why centering these?

Haha yea I realized that all of them are not centered... They probably should be though, we might've broken something at some point.

If you can feel free to center them all the same way :)

app/javascript/controllers/theme_controller.js
kreativmonkey: I have placed the bootstrap icons in the folder so they are now path based. Bootstrap Icons are under MIT licence so they can be used for the project. I also added informations about these icons in the theme/LICENSE.md

Robin481 (approved):

PR #1203, opened 2026-06-23
Bug/1153 Automatic Updates shouldn't interfere with reminder mail

### LINKED ISSUE #1153: Automatic Updates shouldn't interfere with reminder mail
When things like PTime sync updated an user the updated_at timestamp gets updated. This kind of ruins our reminder_mail as its hard for a profile to not get any automatic updates for 6 months.

**TODO**
- [ ] Make sure logins don't trigger auto updates
- [ ] Check if its still an issue or only happened because of the reorg
- [ ] Find a solution

schiltpuzzle:


schiltpuzzle:
# PR Status — 23.06.2026

## Done
- The feature is implemented. Needs a Peer Review.

## ToDos
- [ ] Review

Robin481 (changes requested):
Huh doesn't this ignore the most important part of the issue?

You changed it so that the login doesn't cause the person to be updated anymore. That is perfectly fine.

However the whole ptime-sync side, which was the original reason why the issue was created, is ignored.
Currently any change in pTime (such as the departments were) causes all profiles to be updated and not receive any email reminders. That whole part should not be ignored and properly be attributed for. 

Granted the issue's TODO list is not really useful on that front. @nevio18324 @schiltpuzzle please think this one through again.

nevio18324:
We decided that this should be enough because of how rarely a Member gets changed inside PTime

Robin481 (approved):

PR #1207, opened 2026-06-30
fix: params getting lost between requests #1097

### LINKED ISSUE #1097: Skillset sorting does not work correctly when switching between default/new/all skills
**Description**
Inside the Skillset tab when sorting after members and then switching from all skills to only default skills the sorting is not correct 

https://github.com/user-attachments/assets/492d3050-6e49-4bec-b979-d9c2169f640d

**TODO**
- [ ] Fix the Bug
- [ ] Tests

nevio18324:


schiltpuzzle (commented):
Looks good to me 👍

Robin481 (changes requested):
If we are already at it, make the default sorting A-Z instead of Z-A 😬

Robin481 (approved):
🥛

PR #1210, opened 2026-07-01
Feature/1205 Security Headers 

### LINKED ISSUE #1205: set content security and cors headers
The configs already exists `config/initializers/cors.rb` and `config/initializers/content_security_policy`
Use [grafana](https://grafana.puzzle.ch/d/Jeb97eCZzR/tls-and-security-header-scan?orgId=1&from=now-14d&to=now&timezone=Europe/Zurich&var-team=$__all&var-host_filter=skills.puzzle.ch) to find out what exactly is missing,

>[!Important]
>Make sure images from `puzzle.ch` and `ENV[:PICTURE_HOST]` are allowed

**TODO**
- [x] Try to fix the problems [grafana](https://grafana.puzzle.ch/d/Jeb97eCZzR/tls-and-security-header-scan?orgId=1&from=now-14d&to=now&timezone=Europe/Zurich&var-team=$__all&var-host_filter=skills.puzzle.ch) lists
- [x] Update CHANGELOG.md if feature is major
- [x] Update Readme.md if necessary

schiltpuzzle:


schiltpuzzle:
## The Problem 

We have lots of code that uses inline `js` and `css`. The code does't always come from us. Most of the inline code comes from Libraries like slim-select. When Setting the value for script_src and style_src in the Content-Security-Policy to `self` most of the Code from us and the Libraries doesn't Work any more. And Setting the src to unsafe-inline as the name already says is unsafe.  
  
## The Risk in Rails  
  
Rails has a build in XSS Prevention. When you want to inject a script (`<script>alert("Test")</script>`) Rails would handle this as normal text before sending it to a other user. Rails removes/resolves the script or style tags and replaces them with harmless HTML tags.  
  
## General Risk with `style_src: unsafe-inline`  
  
Style-src: unsafe-inline is not quite as vulnerable has script-src: unsafe-inline, but with CSS injection you could:  
  
- They could turn your page pink, and make it look silly.  
- They could modify the text of your page, making it look like you're saying something offensive that could offend your readership audience.  
- They could make user generated content, like a link they provided appear outside of the normal places where people expect to see user content, making it appear official. (eg, replacing a "Login" button on your site with their own link).  
- Using a carefully crafted style rules they could send any information included on the page to external domains and expose or otherwise use that data maliciously against your users.  
  
## Possible Solution  
  
### CSS  
  
CSS is not that big of a problem for us, because we basically don't use inline `css`, because we use Bootstrap or an external style sheet. Setting `CSS` to unsafe inline is not a big vulnerability and I have seen in most websites. It is worth it to leave it at unsafe inline, because most of the time it's not worth the headache.  
  
### Unsafe inline 

When setting unsafe inline for `js`, Ruby on Rails covers 90% of the cases with its built-in `XSS` prevention. Setting unsafe inline would be a valid choice and would be relatively secure at a very low effort.  
  
### Nonce Flow 

You can create a nonce and set that nonce for your `js`. The nonce gets created by rails, that makes it hard to get the token to the frontend. An option for fronted libraries like Slim Select is that we can send the nonce to that library and the library sets the nonce for its assets. But you can do this with only some of the libraries. The other way would be to use unsafe hashes for the libraries.  
  
### Unsafe Hashes  
  
For every violation of the `csp` header, the browser creates a error message. In that message there is a generated hash of the script. That hash can be added to the `csp` header. This also needs the header `unsafe-hashes` to work. This works maybe for some java-script like libraries that do not often change, because the hash is generated on the script that was blocked by the `csp`. I only see this approach work in combination with a nonce flow. Nevertheless, this is very hard to maintain, because every time we deploy, or a library gets and updates, we have to update the hashes.  
  
## Conclusion  
  
I ran the Grafana test locally with unsafe hashes for both `js` and `css` and i got a score around 80/100. I think the unsafe hashes are the best value for its price and the other methods are lots of work only to cover that 10% of the `XSS` cases.

nevio18324:
Seems reasonable to me :+1:  considering our application should generally be protected by keycloak as well.

config/initializers/content_security_policy.rb
```diff
@@ -4,26 +4,30 @@
...
+  config.content_security_policy do |policy|
+    policy.default_src :self, :https
+    policy.font_src    :self, :https, :data
+    policy.img_src     :self, :data, "https://www.puzzle.ch", "*.puzzle.ch"
+    policy.object_src  :none
+    policy.script_src  :self, :unsafe_inline, :https
+    policy.style_src   :self, :unsafe_inline, :https
+
+    policy.connect_src :self, :https, "sso.puzzle.ch"
+    policy.frame_src   :self, :https, "sso.puzzle.ch"
```
Robin481: Probably not needed

config/initializers/content_security_policy.rb
```diff
@@ -4,26 +4,30 @@
...
+
+    # Specify URI for violation reports
+    # policy.report_uri "/csp-violation-report-endpoint"
```
Robin481: Is deprecated anways, remove comment

config/initializers/cors.rb
```diff
@@ -5,12 +5,15 @@
...
-#   allow do
-#     origins "example.com"
-#
-#     resource "*",
-#       headers: :any,
-#       methods: [:get, :post, :put, :patch, :delete, :options, :head]
-#   end
-# end
+Rails.application.config.middleware.insert_before 0, Rack::Cors do
+  allow do
```
Robin481: Probably needs puzzle image sources

config/initializers/cors.rb
```diff
@@ -5,12 +5,15 @@
...
+    origins "http://localhost:3000",
```
Robin481: put as env variables

config/initializers/content_security_policy.rb
```diff
@@ -4,26 +4,30 @@
...
-#   # config.content_security_policy_nonce_auto = true
-#
-#   # Report violations without enforcing the policy.
-#   # config.content_security_policy_report_only = true
-# end
+Rails.application.configure do
+  config.content_security_policy do |policy|
+    policy.default_src :self, :https
+    policy.font_src    :self, :https, :data
+    policy.img_src     :self, :data, "https://www.puzzle.ch", "*.puzzle.ch"
```
Robin481: env variables

config/initializers/content_security_policy.rb
```diff
@@ -4,26 +4,30 @@
...
+    policy.object_src  :none
+    policy.script_src  :self, :unsafe_inline, :https
+    policy.style_src   :self, :unsafe_inline, :https
+
+    policy.connect_src :self, :https, "sso.puzzle.ch"
```
Robin481: env variables

config/initializers/content_security_policy.rb
Robin481: env variables

config/application.rb
```diff
@@ -54,5 +54,12 @@ class Application < Rails::Application
 
     # Prevent health checks from clogging up the logs.
     config.silence_healthcheck_path = "/status/health"
+
+    # Add Cors isolation to default headers
+    config.action_dispatch.default_headers.merge!(
```
Robin481, line 59: I have no idea what this does 😬 please explain

config/environments/production.rb
```diff
@@ -31,7 +31,7 @@
   config.assume_ssl = false
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = false
+  config.force_ssl = true
```
Robin481: This doesn't work as we terminate ssl before our puma process

Robin481 (changes requested):
🧑‍🚀

config/application.rb
schiltpuzzle: - Cross-Origin Resource Policy: Blocks other sites from embedding your files.
- Cross-Origin Opener Policy: Protects your window from malicious popups/openers.
- Cross-Origin Embedder Policy: Forces your page to only load verified third-party files.

https://www.publisher-collective.com/blog/a-simple-guide-to-coop-coep-corp-and-cors

config/initializers/cors.rb
schiltpuzzle: Image sources are not needed in CORS for images to work...

schiltpuzzle:
I removed the cors, because skills runs server-sided. That means the resources are not being called from a different origin. By removing the cors and using the default values, only the same origin can now call the resource.

config/docker/test/home/rails/.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Robin481: Why suddenly this file, seems like a prime candidate for a .gitignore entry

config/initializers/content_security_policy.rb
```diff
@@ -4,26 +4,26 @@
...
-#   # config.content_security_policy_nonce_auto = true
-#
-#   # Report violations without enforcing the policy.
-#   # config.content_security_policy_report_only = true
-# end
+Rails.application.configure do
+  config.content_security_policy do |policy|
+    policy.default_src :self, :https
+    policy.font_src    :self, :https, :data
+    policy.img_src     :self, :data, ENV.fetch('PICTURE_HOST', 'puzzle.ch')
```
Robin481: I wouldn't even set a default here. Just use the ENV variable

config/initializers/content_security_policy.rb
```diff
@@ -4,26 +4,26 @@
...
+    policy.object_src  :none
+    policy.script_src  :self, :unsafe_inline, :https
+    policy.style_src   :self, :unsafe_inline, :https
+
+    policy.connect_src :self, :https, ENV.fetch('CONNECT_SRC', "")
```
Robin481, line 16: Does the fallback `""` make sense or can this accidentally be too permissive?

Robin481 (changes requested):

config/initializers/content_security_policy.rb
schiltpuzzle: A empty string will not be added to the CSP, that means the `connect_src` will fallback to `:self`

Robin481 (approved):

PR #1211, opened 2026-07-01
feature/1151 add docx templates

### LINKED ISSUE #1151: Recreate all document templates in `.docx`
We need to recreate every existing template option in `.docx` format to prepare for future export functionality. These templates should mirror the current `.odt` versions as closely as possible, using the `.odt` files as the source of truth
Some `.odt` formatting attributes may not be supported or may behave differently in `.docx`, so you need to recreate them from hand.

**TODO**
- [ ] Every current `.odt` template has an equivalent `.docx` version
- [ ] Test outside browser-based Word environments
- [ ] Tests it outside of browser version of word.
- [ ] Talk with @Robin481 or someone of marcom about the styling.

IrynaKosiak7:
fixed #1151


nevio18324 (approved):
The Templates look nice 👍 The only thing that would be nice if you could rename your branch to feature/1151-add-docx-templates

and for the next time can you write your commit messages using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

PR #1215, opened 2026-07-07
feat: allow user to create skills inside dropdown #1112

### LINKED ISSUE #1112: Implement create-skill button in skill-drop-down
**Description**
First of all, take a look at this [document](https://codimd.puzzle.ch/lyX_PlJoQkeihCjFhCNmng?both#). What’s in the document is the goal, but this ticket only requires the implementation of the create new skill button. This button appears when you search for a skill that isn't available. When you click the button, the modal window that we already need to create a new skill should appear. In the modal, the title should already be filled in with whatever you entered in the drop-down search bar. When a new skill has been created in the modal, the modal should close and the newly created skill should be added to the drop-down menu

**Additional context**
***Mock-up***

Image

[Mock-ups](https://files.puzzle.ch/apps/collectives/BBT-Java-24/UI-8906767)

**TODO**
- [ ] create new skill button appears when no skill is found
- [ ] When clicking the add button the add-skill-modal opens with filled in title
- [ ] Once created, the skill should be added to the drop-down menu 
- [ ] Test

**Dependence**
- #1111

nevio18324:


schiltpuzzle (commented):
The code looks good, but I found some issues will testing:

- When clicking on the new button "The Skill ... doesn't exist yet" and then going back one page with the browser function, the drop-down becomes unusable.
- When hovering over the link, the contrast gets really bad. A option is to remove the hover button effect and display a native link.
- This is more of a personal opinion: I think I would help with usability by adding a plus icon as a prefix for the link.

Robin481 (changes requested):
Before I review the code, the idea is solid but needs a lot of usability improvements:
- [x] Do not make the link to the skillset open in the same tab. While it works that you can go back and the browser keeps the inputs, the user might not realize this and fear that he has lost his inputs.
- [x] Adding to the above, even if this was okay. After creating the skill and using browser back to navigate back to the project the newly created skill won't show up at all. You have to reload the browser page so that the skill shows up losing all of your inputs anyway. The dropdown needs to refresh in some form to let newly created skills show up.
- [x] The option to create a new skill doesn't highlight when navigating onto it with they keyboard. However you are actually focused onto it and pressing enter will open it. Very confusing.
- [x] The dropdown doesn't find skills with multiple words at all. Example: `ruby on rails`: it only finds it when searching for `ru` after that when entering `rub` it will be stuck `searching...` which is fine for a second but should never get stuck. Also it should just find multiple word skills 😬

schiltpuzzle:
Code looks good and everything work, expect

- [x] When searching through the option with the input it gets stuck when you type more than 2 characters and displays "searching..."

nevio18324:
> Code looks good and everything work, expect
> 
>  When searching through the option with the input it gets stuck when you type more than 2 characters and displays "searching..."

This seems to be a issue on slimselects side. I created a issue so it maybe gets fixed.
https://github.com/brianvoe/slim-select/issues/695

nevio18324:
The Issue was fixed by slim select

app/controllers/skills_controller.rb
```diff
@@ -4,14 +4,17 @@ class SkillsController < CrudController
...
 
   helper_method :check_entry_validity?
 
   self.permitted_attrs = %i[title radar portfolio default_set category_id]
 
   def index
     entries
+    if params[:open_modal]
+      @skill = Skill.new(params[:skill])
+    end
```
Kagemaru: Modifikatoren ftw:
```suggestion
    @skill = Skill.new(params[:skill]) if params[:open_modal]
```

Kagemaru (changes requested):
Skill search is broken since this change, pls fix!

Kagemaru (approved):
Works now, thanks. Remember to recreate your assets!

PR #1216, opened 2026-07-08
Bug/1108 Incorrect green error message shown for duplicate title in Unify Skills

### LINKED ISSUE #1108: Incorrect green error message shown for duplicate title in Unify Skills
When attempting to unify two skills using a title that already exists, a success-style message is displayed in green.

This is misleading, as green typically indicates a successful operation. In this case, the action does not complete successfully, and the user is incorrectly led to believe that the operation worked.

**Additional context**
Context

**TODO**
- [ ] Change it to red
- [ ] Tests 👷🏻‍♂️

lcanobbio:


schiltpuzzle (commented):
The code looks good, BUT ...

- At the moment you are checking case-sensitive, that means the category will not be shown when a skill `bash` and a skill `Bash` exist.
- When you go to the skills tab of a person and want to add a new skill, the categories are missing from the drop-down.
- Lots of skills on prod use brackets as well for displaying extra information, maybe signaling more that the categories are not part of the name.

app/views/skill_search/_filter_row.html.haml
```diff
@@ -2,7 +2,13 @@
   %div.filter-container
     %div.w-100.d-flex.flex-row.justify-content-start{data: {controller: "skills-level"}}
       %div.w-50.align-self-end{"data-controller": "dropdown"}
-        = form.select "skill_id[]", skills_for_select(filter.skill_id), { prompt: true }, data:{"dropdown-target": "dropdown"}, onchange: "this.form.requestSubmit()"
+        - disabled_skills = @search.filter_rows.map(&:skill_id).excluding(filter.skill_id)
```
Robin481: Put into controller not here

app/views/skill_search/_filter_row.html.haml
```diff
@@ -2,7 +2,13 @@
...
+        - options = skills_for_select(filter.skill_id, disabled_skills: disabled_skills)
```
Robin481: Put into controller not here

Robin481 (changes requested):

app/views/skill_search/_filter_row.html.haml
lcanobbio: I moved it to a helper because we need to pass in the current filter to see which skills are already disabled, and we can't do that in the controller.

app/views/skill_search/_filter_row.html.haml
lcanobbio: I moved it to a helper because we need to pass in the current filter to see which skills are already disabled, and we can't do that in the controller.

Robin481 (changes requested):
It's really nice that you display the category when there is a duplicate.
But please also do this in the unify skills mode itself as well 😬

Robin481 (approved):
☄️

PR #1227, opened 2026-07-15
Chore/1223 bump rails and gems version

### LINKED ISSUE #1223: Bump Rails, Ruby and Gems
Title says it all

schiltpuzzle:


app/controllers/admin/unified_skills_controller.rb
```diff
@@ -83,10 +83,10 @@ def update_people_skills(new_skill_id)
...
                .each_value do |people_skills|
 
-                 best_people_skill = people_skills.max_by { |ps| ps.level.to_i }
-                 other_people_skills = people_skills - [best_people_skill]
-                 best_people_skill&.update!(skill_id: new_skill_id)
-                 other_people_skills.each { |ps| ps.delete }
+      best_people_skill = people_skills.max_by { |ps| ps.level.to_i }
+      other_people_skills = people_skills - [best_people_skill]
+      best_people_skill&.update!(skill_id: new_skill_id)
+      other_people_skills.each { |ps| ps.delete }
```
nevio18324, line 89: weird indentation

app/helpers/ptime_helper.rb
```diff
@@ -22,9 +22,9 @@ def provider_configs
        .to_h
        .each_slice(4)
        .map do |config|
-         config.to_h.transform_keys do |key|
-           key.sub(/^PTIME_PROVIDER_[0-9]+_/, '')
-         end
+      config.to_h.transform_keys do |key|
+        key.sub(/^PTIME_PROVIDER_[0-9]+_/, '')
+      end
```
nevio18324, line 27: same here

config/initializers/new_framework_defaults_8_1.rb
nevio18324, line 1: you can delete this file

config/puma.rb
```diff
@@ -25,9 +25,8 @@
 # Any libraries that use a connection pool or another resource pool should
 # be configured to provide at least as many connections as the number of
 # threads. This includes Active Record's `pool` parameter in `database.yml`.
-max_threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }
-min_threads_count = ENV.fetch("RAILS_MIN_THREADS") { max_threads_count }
-threads min_threads_count, max_threads_count
+threads_count = ENV.fetch("RAILS_MAX_THREADS", 5)
+threads threads_count, threads_count
```
nevio18324: I think we lost our min threads

Gemfile
```diff
@@ -58,7 +58,7 @@ gem 'turbo-rails'
 group :metrics do
   gem 'brakeman'
   gem 'haml_lint', require: false
-  gem 'rubocop', '1.84.1'
+  gem 'rubocop', '>= 1.88.2'
```
nevio18324: you can just remove the version specification

Gemfile
```diff
@@ -92,7 +92,7 @@ end
 
 group :test do
   gem 'capybara'
-  gem 'selenium-webdriver', '>= 4.28.0'
+  gem 'selenium-webdriver', '>= 4.46'
```
nevio18324: same here

nevio18324 (changes requested):
I have noticed a lot of indentations are looking kinda odd .
Is rubocop doing that?

config/puma.rb
schiltpuzzle: Yeah that is the whole point. Setting the min equal to the max is best practice. This ensures that the resource usage stays the same over time.

app/controllers/admin/unified_skills_controller.rb
```diff
@@ -83,7 +83,7 @@ def update_people_skills(new_skill_id)
                .group_by(&:person_id)
                .each_value do |people_skills|
 
-                 best_people_skill = people_skills.max_by { |ps| ps.level.to_i }
+                 best_people_skill = people_skills.max_by { |ps| ps.level.to_i } # rubocop:disable Layout/IndentationWidth
```
Robin481: Can't you find a way to do this without disabling the rule?

app/helpers/ptime_helper.rb
```diff
@@ -22,7 +22,7 @@ def provider_configs
        .to_h
        .each_slice(4)
        .map do |config|
-         config.to_h.transform_keys do |key|
+         config.to_h.transform_keys do |key| # rubocop:disable Layout/IndentationWidth
```
Robin481: again fix this without disabling rubocop

.rubocop.yml
```diff
@@ -182,4 +182,7 @@ Style/Attr:
   Enabled: false
 
 CustomCops/TranslatedHamlFiles:
-  Enabled: true
\ No newline at end of file
+  Enabled: true
+
+Rails/StrongParametersExpect:
+  Enabled: false
```
Robin481, line 188: Why don't we enable this?

Robin481 (changes requested):

.rubocop.yml
schiltpuzzle: This forces us to use params.expect(), which throws an error when the param doesn't exist. At the moment, we use params[], which returns nil when the param is not found. Throwing an error instead of returning nil breaks our app. I discussed this with @nevio18324, and we came to the conclusion that a refactor isn't worth it because the new method doesn't bring any major improvements to the table.

app/controllers/admin/unified_skills_controller.rb
schiltpuzzle: First i had `rubocop` fix all the indentation errors, but @nevio18324 disproved with them, because they where out of place.

I fixed them in the previous commit. You can see for yourself and decide what you think about the `rubocop` fix:

https://github.com/puzzle/skills/pull/1227/changes/2eff1a60dca4c9e6dc100a72423af29a40bf3e93

.rubocop.yml
Robin481: Create a follow up ticket so we can refactor and enable this.

Robin481 (approved):

PR #1238, opened 2026-07-17
Picture host handling improved

schiltpuzzle:
This PR fixes that the validation for our picture url and the `img_src` have other syntax for the `PICTURE_HOST` env.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -138,13 +138,21 @@ class Person < ApplicationRecord
     url = URI.parse(picture_url)
     return false if url.host.nil?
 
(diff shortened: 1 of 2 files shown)


schiltpuzzle:
# PR Status — 17.07.2026

## Done
- I refactored the picture validation so it has the same syntax as the CSp Config. I removed the default form the Picture Validation. That means the tests are failing. To set the env Variable for rspec you have to write it into the `rails_helper.rb`. This didn't fix the test for me, maybe there is more to it :scream:  ...

## ToDos
- [ ] Fix the test

